PR #14, opened 2022-02-04
Avoid calling `buildFinished` after build is finished

### LINKED ISSUE #4: Calling buildFinished when build is finished is too late
When asynchronously invoking code via buildScan.background, it can happen that this background task finishes _after_ the build has finished. The GE plugin will correctly wait for the background task to finish. The problem in the CCUD plugin is that, currently, the background task invoked from `captureGitMetadata` will add a custom link via buildFinished call - which is too late when the end of the build has already been reached since the GE plugin will no longer allow to add more buildFinished listeners once the build has finished. This can be observed when a warning is printed by the GE plugin at the ened of the build. As explained, this does not happen everytime but under certain conditions.

It seems to me that we should go with the approach we had before: having a list and adding the actions there (in a thread-safe way) and executing them by a single buildFinished listener. Of course we need to be cautious and avoid an infinite loop where an action adds another action.

bigdaz:
Search links for custom values must be added during a 'buildFinished' event,
in order to ensure that the server URL is set to it's final value.
The previous implementation did this by registering a separate 'buildFinished'
listener for each search link, but the background process that captured git metadata
would occasionally register after the build was complete.

In order to avoid the warnings issued in this case, we now register a single
buildFinished listener, with a thread-safe implementation that will collect all
registered build scans. With this change, if the git metadata process completes after
the build is complete, then the search link is ignored and no warning is issued.

Note that configuration-cache compliance necessitated the use of a static class
to register as a buildFinished listener.

Fixes #4

### DIFF
--- a/src/main/java/com/gradle/CustomBuildScanEnhancements.java
+++ b/src/main/java/com/gradle/CustomBuildScanEnhancements.java
@@ -1,5 +1,6 @@
 package com.gradle;
 
+import com.gradle.scan.plugin.BuildResult;
 import com.gradle.scan.plugin.BuildScanExtension;
 import org.gradle.api.Action;
 import org.gradle.api.Task;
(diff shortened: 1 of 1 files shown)


runningcode (approved):
LGTM! Thanks

bigdaz:
OK cool. I did try further refactoring but kept hitting issues with configuration-cache compatibility. Glad you found a way to get it working.

PR #22, opened 2022-03-15
Fix missing signature files when publishing to the plugin portal

alextu:
`plugin-publish` Gradle plugin starting from version 0.14.0 does not publish the signature files when used with the `maven-publish` Gradle plugin.

Until this bug is fixed, this PR:
- keeps the `maven-publish` plugin so users can still build and publish to their internal repository
- configures the `plugin-publish` so it ignores the `maven-publish` model and publish the signatures

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -24,6 +24,8 @@ java {
 }
 
 gradlePlugin {
+    automatedPublishing = false
+
     plugins {
         commonCustomUserData {
             id = 'com.gradle.common-custom-user-data-gradle-plugin'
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -49,6 +50,10 @@ pluginBundle {
     website = 'https://github.com/gradle/common-custom-user-data-gradle-plugin'
     vcsUrl = 'https://github.com/gradle/common-custom-user-data-gradle-plugin.git'
     tags = ['android', 'java', 'gradle enterprise']
+    mavenCoordinates {
```
etiennestuder, line 53: Please add a blank line above.

build.gradle
```diff
@@ -24,6 +24,7 @@ java {
 }
 
 gradlePlugin {
+    automatedPublishing = false
     plugins {
```
etiennestuder, line 28: Please add a blank line above.

etiennestuder (approved):
Just some formatting stuff. I cannot comment on the fix itself, but I assume this is the outcome of the research and that it is all good.

@alextu can you please do the same improvement to the cache fix plugin and to the wrapper upgrade plugin?

etiennestuder:
Once this is in, it will be worth to release a new patch version.

alextu:
> Just some formatting stuff. I cannot comment on the fix itself, but I assume this is the outcome of the research and that it is all good.
> 
> @alextu can you please do the same improvement to the cache fix plugin and to the wrapper upgrade plugin?

Yes will do after releasing a patch of this one

PR #36, opened 2022-04-08
Enable remote build caching

clayburn:
This change enables the remote build cache for all builds. Anonymous
users have read-only build cache access, while CI has read-write.

This change also configures GitHub actions to authenticate with the
build cache.


etiennestuder (approved):
Looks good. Nothing to improve here. Please merge.

PR #38, opened 2022-04-12
Add functional test for configuration cache

clayburn:


build.gradle
```diff
@@ -50,6 +53,37 @@ tasks.withType(ValidatePlugins).configureEach {
     enableStricterValidation = true
 }
 
+// A local repo we publish our library to for testing in order to workaround limitations
+// in the TestKit plugin classpath.
+def localRepo = file("$buildDir/local-repo")
+publishing {
+    // Since for now we need to set `gradlePlugin { automatedPublishing = false ... }`, we have to create manually
```
runningcode, line 60: i think lines 60-71 are not required to get this working (to give you a better idea of what is happening here)

build.gradle
```diff
@@ -50,6 +53,37 @@ tasks.withType(ValidatePlugins).configureEach {
...
+            url = localRepo.toURI()
+        }
+    }
+}
+
+tasks.withType(Test).configureEach {
+    dependsOn publish
+    workingDir projectDir
+    systemProperty "local.repo", projectDir.toPath().relativize(localRepo.toPath()).toString()
+    systemProperty "com.gradle.common-custom-user-data-gradle-plugin.version", version
```
runningcode, line 83: is this version property used somewhere?

runningcode, line 83: is this version property used somewhere?

build.gradle
```diff
@@ -50,6 +53,37 @@ tasks.withType(ValidatePlugins).configureEach {
...
+    repositories {
+        maven {
+            url = localRepo.toURI()
+        }
+    }
+}
+
+tasks.withType(Test).configureEach {
+    dependsOn publish
+    workingDir projectDir
```
runningcode, line 81: is this required?

runningcode, line 81: is this required?

build.gradle
```diff
@@ -50,6 +53,37 @@ tasks.withType(ValidatePlugins).configureEach {
...
+    systemProperty "local.repo", projectDir.toPath().relativize(localRepo.toPath()).toString()
+    systemProperty "com.gradle.common-custom-user-data-gradle-plugin.version", version
+    useJUnitPlatform()
```
runningcode, line 84: i don't think this is needed either.

runningcode, line 84: i don't think this is needed either.

src/test/groovy/com/gradle/ConfigurationCachingFuncTest.groovy
```diff
@@ -0,0 +1,125 @@
...
+
+import org.gradle.testkit.runner.GradleRunner
+import spock.lang.Specification
+import spock.lang.TempDir
+
+import java.nio.file.Paths
+
+class ConfigurationCachingFuncTest extends Specification {
+    static final String CC_PROBLEMS_FOUND = "problems were found storing the configuration cache"
+    static final String PLUGIN_VERSION_SYSTEM_PROPERTY = 'com.gradle.common-custom-user-data-gradle-plugin.version'
```
runningcode, line 11: i think we only want this if we're doing cross version testing which I don't think we're doing here.

runningcode, line 11: i think we only want this if we're doing cross version testing which I don't think we're doing here.

runningcode (commented):
@clayburn I just added some comments that might help simplify this.

In general: i think this is too complicated to add to the CCUD plugin if we want customers to be able to understand this build but I think it would be a good excercise to boil this change down to the essentials by simplifying it.

PR #47, opened 2022-05-19
Proxy Gradle Enterprise plugin types to work around ClassLoader issues

marcphilipp:
When the GE plugin is applied by the build and the CCUD plugin via an
init script, the Gradle Enterprise extension and types it references
are loaded by a different classloader than the classes in the CCUD
plugin. To work around this issue, the Gradle Enterprise types are now
accessed via dynamic proxies that "translate" between the different
class loaders at runtime.

### DIFF
--- a/src/main/java/com/gradle/CommonCustomUserDataGradlePlugin.java
+++ b/src/main/java/com/gradle/CommonCustomUserDataGradlePlugin.java
@@ -39,7 +39,8 @@ public class CommonCustomUserDataGradlePlugin implements Plugin<Object> {
         settings.getPluginManager().withPlugin("com.gradle.enterprise", __ -> {
             CustomGradleEnterpriseConfig customGradleEnterpriseConfig = new CustomGradleEnterpriseConfig();
 
-            GradleEnterpriseExtension gradleEnterprise = settings.getExtensions().getByType(GradleEnterpriseExtension.class);
+            Object extension = settings.getExtensions().getByName("gradleEnterprise");
+            GradleEnterpriseExtension gradleEnterprise = ProxyFactory.createProxy(extension, GradleEnterpriseExtension.class);
             customGradleEnterpriseConfig.configureGradleEnterprise(gradleEnterprise);
 
             BuildScanExtension buildScan = gradleEnterprise.getBuildScan();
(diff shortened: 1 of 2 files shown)


src/main/java/com/gradle/ProxyFactory.java
```diff
@@ -0,0 +1,108 @@
...
+
+        @Override
+        public Object invoke(Object proxy, Method method, Object[] args) {
+            try {
+                Class<?>[] parameterTypes = method.getParameterTypes();
+                Method targetMethod = target.getClass().getMethod(method.getName(), convertTypes(parameterTypes, target.getClass().getClassLoader()));
+                Object result = targetMethod.invoke(target, toTargetArgs(args));
+                if (result == null || isJdkType(result.getClass())) {
+                    return result;
+                }
```
leonard84, line 37: ❔ Did you decide caching the lookup wasn't worth it?

src/main/java/com/gradle/CommonCustomUserDataGradlePlugin.java
```diff
@@ -39,7 +39,8 @@ private void applySettingsPlugin(Settings settings) {
         settings.getPluginManager().withPlugin("com.gradle.enterprise", __ -> {
             CustomGradleEnterpriseConfig customGradleEnterpriseConfig = new CustomGradleEnterpriseConfig();
 
-            GradleEnterpriseExtension gradleEnterprise = settings.getExtensions().getByType(GradleEnterpriseExtension.class);
+            Object extension = settings.getExtensions().getByName("gradleEnterprise");
+            GradleEnterpriseExtension gradleEnterprise = ProxyFactory.createProxy(extension, GradleEnterpriseExtension.class);
```
leonard84, line 43: ❔ could we check if the classloaders are compatible and only fall back to reflection if they are not?

marcphilipp: I discussed this with Luke and we agreed to always do this in order to get more coverage and spot potential issues sooner.

src/main/java/com/gradle/ProxyFactory.java
marcphilipp: I had a cache in some interim version but removed it to restore configuration cache compatibility. I don't think performance is a real concern here since these APIs are not called excessively.

etiennestuder (approved):
LGTM. I went until `adaptActionArg`.

PR #49, opened 2022-05-19
Changes to make it Gradle 4.x and GE plugin 1.16 compatible

clayburn:
See here for a sample scan where I set some env variables to pretend like it was running on GitHub Actions: https://scans.gradle.com/s/6eeektdxxtpby


src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -400,9 +409,14 @@ public synchronized void execute(BuildResult buildResult) {
...
     private void captureTestParallelization() {
-        gradle.afterProject(p ->
-            p.getTasks().withType(Test.class).configureEach(captureMaxParallelForks(buildScan))
-        );
+        gradle.afterProject(p -> {
+            TaskCollection<Test> tests = p.getTasks().withType(Test.class);
+            if (GradleVersion.current().compareTo(GradleVersion.version("5.0")) >= 0) {
+                tests.configureEach(captureMaxParallelForks(buildScan));
+            } else {
+                tests.all(captureMaxParallelForks(buildScan));
```
clayburn, line 417: I have no idea what the state of Lazy Configuration was in 4.x, so `all` may not be best here.

etiennestuder (approved):
LGTM. I will do a bit of polishing post-merge.

src/main/java/com/gradle/CommonCustomUserDataGradlePlugin.java
```diff
@@ -88,8 +92,36 @@ private void applyProjectPlugin(Project project) {
...
 
+    private void applyProjectPluginGradle4(Project project) {
+        if (!project.equals(project.getRootProject())) {
+            throw new GradleException("Common custom user data plugin may only be applied to root project");
+        }
+        project.getPluginManager().withPlugin("com.gradle.build-scan", __ -> {
+            CustomGradleEnterpriseConfig customGradleEnterpriseConfig = new CustomGradleEnterpriseConfig();
+
+            BuildScanExtension buildScan = project.getExtensions().getByType(BuildScanExtension.class);
+            customGradleEnterpriseConfig.configureGradleEnterprise(buildScan);
```
etiennestuder, line 103: Getting by class name will not work in all cases, we need to get by name and use the proxy.

src/main/java/com/gradle/CustomBuildScanEnhancements.java
etiennestuder: LGTM

src/main/java/com/gradle/CommonCustomUserDataGradlePlugin.java
etiennestuder: Changed.

PR #56, opened 2022-06-01
Fix Allow Untrusted Server override

clayburn:


bigdaz:
@clayburn LGTM. 

Note that on Gradle the same bug causes the `gradle.enterprise.allowUntrustedServers` property to be silently ignored: no exception is thrown.

etiennestuder (approved):
LGTM

PR #58, opened 2022-06-02
Apply Overrides outside of `settingsEvaluated` when executed in a `settingsEvaluated` callback

clayburn:
This works around the issue where Override do not work when CCUD is applied in a `settingsEvaluated` callback


etiennestuder (approved):

etiennestuder:
Thanks, @clayburn

Issue #74, opened 2022-07-12
add BRANCH_NAME even if git is missing


NikolayMetchev:
We are using the k8s plugin in Jenkins which runs gradle in a docker container that doesn't have git installed. This means that the BRANCH name is no longer added as a tag. It would be good if this was fixed so that even if git is not installed the BRANCH_NAME env variable is still added as a tag.

runningcode:
This is the default behavior. [See the code here](https://github.com/gradle/common-custom-user-data-gradle-plugin/blob/main/src/main/java/com/gradle/CustomBuildScanEnhancements.java#L337)

However in your specific case, since you are running the build inside the docker container, can you make sure that these environment variables are also available inside your docker container?

runningcode:
Ah, thanks for finding that. We'll look in to this.

PR #76, opened 2022-07-15
Get branch name from env var (#74)

alextu:
The implemented behavior will try to get the branch name from the env var first (even if the Git CLI is available).
This PR includes some refactoring to rely on Optional.


src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -281,44 +278,35 @@ private CaptureGitMetadataAction(ProviderFactory providers, CustomValueSearchLin
...
                 buildScan.tag("Dirty");
-                buildScan.value("Git status", gitStatus);
-            }
+                buildScan.value("Git status", s);
+            });
 
-            if (isNotEmpty(gitRepo) && isNotEmpty(gitCommitId)) {
-                if (gitRepo.contains("github.com/") || gitRepo.contains("github.com:")) {
-                    Matcher matcher = Pattern.compile("(.*)github\\.com[/|:](.*)").matcher(gitRepo);
+            if (gitRepo.isPresent() && gitCommitId.isPresent()) {
```
runningcode, line 300: the previous logic would also ensure the resolved value is not empty, could we also check that here?

src/main/java/com/gradle/GitMetadata.java
```diff
@@ -0,0 +1,42 @@
...
+
+    private GitMetadata(boolean isGitInstalled) {
+        this.isGitInstalled = isGitInstalled;
+    }
+
+    static GitMetadata given(boolean isGitInstalled) {
+        return new GitMetadata(isGitInstalled);
+    }
+
+    GitMetadata fromCmd(String... args) {
```
runningcode, line 22: it seems that this class is very close to the standard Java Builder pattern. the `fromCmd` is also a required parameter. can we adjust this class to use the standard Java Builder pattern where this is a requirement?

src/main/java/com/gradle/GitMetadata.java
```diff
@@ -0,0 +1,42 @@
...
+
+    private String[] args;
+    boolean isGitInstalled;
+    private Supplier<Optional<String>> fromEnv;
+
+    private GitMetadata(boolean isGitInstalled) {
+        this.isGitInstalled = isGitInstalled;
+    }
+
+    static GitMetadata given(boolean isGitInstalled) {
```
runningcode, line 18: i'm not a fan of the name `given` here for the static constructor method since it isn't very descriptive when used inline.

src/main/java/com/gradle/GitMetadata.java
```diff
@@ -0,0 +1,42 @@
+package com.gradle;
+
+import java.util.Optional;
+import java.util.function.Supplier;
+
+import static com.gradle.Utils.execAndGetStdOut;
+
+class GitMetadata {
```
runningcode, line 8: I like having a class here encapsulate the logic for fetching the gitmetadata but it seems to me that the encapsulation is not fully complete and requires the caller to know the implementation details. for example you must know you need to pass either the `cmd` args or the value supplier depending on whether git is installed or not.

runningcode (commented):
looks good, i have mostly comments around the `GitMetadata` class but I can't think of the best way to improve it in my head.

src/main/java/com/gradle/CustomBuildScanEnhancements.java
alextu: I could but the current code implicitly already handles the empty strings in the nested ifs

src/main/java/com/gradle/GitMetadata.java
alextu: good point, see latest changes, I think it looks better. There's a hidden trap there: since I keep the builder around for simplicity sake, you have to be careful not to re-use the previous state. On the other hand re-creating the builder every time seems overkill...

src/main/java/com/gradle/GitMetadata.java
alextu: That's a good point, actually I made a change to be more flexible and silently accept the case: git is installed but no cmd is provided.

src/main/java/com/gradle/GitMetadata.java
```diff
@@ -0,0 +1,54 @@
...
+        if (fromEnv != null) {
+            return fromEnv.get();
+        }
+        if (isGitInstalled && args != null) {
+            return execAndGetStdOut(args);
+        }
+        return Optional.empty();
+    }
+
+    static class GitMetadataBuilder {
```
runningcode, line 30: in the current state, this feels a bit more like a `Factory` to me :D 

I'm trying to think of the best way to structure this though. i haven't really fully thought it out but the current implementation seems like it could be simplified.

runningcode: yeah, i think a factory would make sense since it would remove the case of the `.build().resolve()` that is used in every case.

alextu: How about the latest refactoring a bit less expressive but simple: a resolver that just resolves :))

src/main/java/com/gradle/CustomBuildScanEnhancements.java
alextu: I've hardened this by making sure having an empty string does not happen (see GitMetadataResolver logic and test)

runningcode (commented):
Looks good to me aside from the testing change.

We have intentionally kept this repository free of tests in order to make it as simple as possible for customers to copy and understand. I am not saying we should keep or remove the tests but I think it needs to be discussed before making a change like this to add these.

@etiennestuder also wants to take a look at this so I'll let him have the final review.

erichaagdev (approved):

etiennestuder:
I would hold off on merging this PR until we have resolved the config caching issue explored in [this PR](https://github.com/gradle/common-custom-user-data-gradle-plugin/pull/116) and raised in [this issue](https://github.com/gradle/common-custom-user-data-gradle-plugin/issues/80).


I believe we need to make significant changes to where we read the env vars in order to not cause constant config cache invalidation. Depending on how that will be solved, it might require a different approach/implementation.

runningcode:
Closing PR for now. Let's re-open after making the CC compatibility changes.

src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -381,24 +369,15 @@ private boolean isGitInstalled() {
             return execAndCheckSuccess("git", "--version");
         }
 
-        private String getGitBranchName(Supplier<String> gitCommand) {
+        private Optional<String> getGitBranchNameFromEnv() {
```
erichaagdev, line 372: Why is GitHub actions not included?

etiennestuder:
@erichaagdev how often does this actually come up? We have 1 case for sure which is the one that triggeed this PR. Before time is spend on this PR, we need to certain is worth the added complexity to itegrate and maintain. 1 single customer would not justify it.

runningcode:
I don't think it was intentionally left out. 
I think we should also include GitHub actions but this PR is very, very old. This would improve our samples and our builds. For example [here](https://ge.solutions-team.gradle.com/s/irwc3blt64vjg#info)

erichaagdev:
@etiennestuder This would address the issue where CI checks out the branch in a detached HEAD state. In this case, 
`git rev-parse --abbrev-ref HEAD` returns `HEAD` rather than the branch name. 

This is the case for GitHub Actions when a workflow is triggered by a `pull_request` event. For example, consider this build for `android-cache-fix-gradle-plugin`:

- GitHub Actions checks out the branch in a detached `HEAD` state: https://github.com/gradle/android-cache-fix-gradle-plugin/actions/runs/4299768295/jobs/7495318117#step:2:372
- The corresponding Build Scan is tagged with `HEAD` and has a custom value of `Git branch=HEAD`: https://ge.solutions-team.gradle.com/s/5kt2y4de3ahka/custom-values#L2

My expectation in this case is that the Build Scan is tagged with `erichaagdev/update-daily-dependabot-time` and has a custom value of `Git branch=erichaagdev/update-daily-dependabot-time`.

Of course this has the side effect that you cannot filter the Build Scan dashboard to include only builds for your branch.

To better understand the impact this has, consider this filter `project: android-cache-fix-gradle-plugin` and `tags: CI not:main not:head` for the last 90 days. There are only 16 builds shown: https://ge.solutions-team.gradle.com/scans?search.relativeStartTime=P90D&search.rootProjectNames=android-cache-fix-gradle-plugin&search.tags=ci%2Cnot:head%2Cnot:main&search.timeZoneId=America/Chicago

Now if we drop the `tag: HEAD`, we have 1954 builds: https://ge.solutions-team.gradle.com/scans?search.relativeStartTime=P90D&search.rootProjectNames=android-cache-fix-gradle-plugin&search.tags=ci,not:main&search.timeZoneId=America/Chicago

This PR also defines a pattern for how to address #114. In that case, builds running on Bamboo with the repository caching feature enabled will see the `Git repository` custom value as a local file path, rather than the remote repository. See the issue for more details.

etiennestuder:
@erichaagdev Thanks for the details. But you are not addressing the core of my question which is: how often does this come up with customers?

erichaagdev:
Sorry, @etiennestuder but I will not be able to provide an exact frequency.

This would be a problem for any customer using GitHub Actions to run workflows on `pull_request` events, or any other CI provider which checks out projects in a detached state.

The similar #114 issue that I referenced is actually from a customer.

runningcode:
Closing in favor of https://github.com/gradle/common-custom-user-data-gradle-plugin/pull/240

PR #84, opened 2022-08-11
Provide static application method

facewindu:
Related to https://github.com/gradle/ge/issues/16749.
See https://github.com/gradle/ge/pull/16816

This changes adds a publicly available **static** `apply` method on the CCUD gradle plugin in order to make its application work in the Gradle Enterprise build.

### DIFF
--- a/src/main/java/com/gradle/CommonCustomUserDataGradlePlugin.java
+++ b/src/main/java/com/gradle/CommonCustomUserDataGradlePlugin.java
@@ -67,7 +67,7 @@ public class CommonCustomUserDataGradlePlugin implements Plugin<Object> {
 
         // configuration changes applied in this block will override earlier configuration settings,
         // including those set in the settings.gradle(.kts)
-        Action<Settings> settingsAction = __ -> {
+        Action<Settings> settingsAction = ___ -> {
             Overrides overrides = new Overrides(providers);
             overrides.configureGradleEnterprise(gradleEnterprise);
             overrides.configureBuildCache(buildCache);
@@ -155,8 +155,8 @@ public class CommonCustomUserDataGradlePlugin implements Plugin<Object> {
 
     private static boolean settingsHaveBeenEvaluated() {
         return Arrays.stream(Thread.currentThread().getStackTrace())
-                .map(StackTraceElement::getMethodName)
-                .anyMatch(s -> s.contains("settingsEvaluated"));
+            .map(StackTraceElement::getMethodName)
+            .anyMatch(s -> s.contains("settingsEvaluated"));
     }
 
 }
--- a/src/main/java/com/gradle/CommonCustomUserDataGradlePlugin.java
+++ b/src/main/java/com/gradle/CommonCustomUserDataGradlePlugin.java
@@ -45,13 +45,14 @@ public class CommonCustomUserDataGradlePlugin implements Plugin<Object> {
 
     public static void apply(Object gradleEnterprise, ProviderFactory providers, Settings settings) {
         applySettingsPlugin(
-                ProxyFactory.createProxy(gradleEnterprise, GradleEnterpriseExtension.class),
+                gradleEnterprise,
                 providers,
                 settings
         );
     }
 
-    private static void applySettingsPlugin(GradleEnterpriseExtension gradleEnterprise, ProviderFactory providers, Settings settings) {
+    private static void applySettingsPlugin(Object gradleEnterpriseExtension, ProviderFactory providers, Settings settings) {
+        GradleEnterpriseExtension gradleEnterprise = ProxyFactory.createProxy(gradleEnterpriseExtension, GradleEnterpriseExtension.class);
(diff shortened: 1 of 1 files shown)


src/main/java/com/gradle/CommonCustomUserDataGradlePlugin.java
```diff
@@ -43,38 +43,46 @@ public void apply(Object target) {
...
+    }
 
-            // configuration changes applied in this block will override earlier configuration settings,
-            // including those set in the settings.gradle(.kts)
-            Action<Settings> settingsAction = ___ -> {
-                Overrides overrides = new Overrides(providers);
-                overrides.configureGradleEnterprise(gradleEnterprise);
-                overrides.configureBuildCache(buildCache);
-            };
+    private static void applySettingsPlugin(GradleEnterpriseExtension gradleEnterprise, ProviderFactory providers, Settings settings) {
```
ldaley, line 54: You could change this method to take `Object gradleEnterprise` and do the proxying internally. That would simplify our use-site, so we won't have to duplicate the proxying.

ldaley (changes requested):

src/main/java/com/gradle/CommonCustomUserDataGradlePlugin.java
ldaley: I've just seen that we don't actually need proxying with our usage currently. I still recommend making this change though in case we want to change things to only load the CCUD plugin once.

facewindu: OK will do
The interesting bit about loading it twice is that we could dogfood dev version of it while keeping a stable version for our stable build scans

facewindu: done

facewindu:
@gradle/ge-solutions can you review that ?

src/main/java/com/gradle/CommonCustomUserDataGradlePlugin.java
```diff
@@ -43,38 +43,47 @@ public void apply(Object target) {
...
+    }
 
-            // configuration changes applied in this block will override earlier configuration settings,
-            // including those set in the settings.gradle(.kts)
-            Action<Settings> settingsAction = ___ -> {
-                Overrides overrides = new Overrides(providers);
-                overrides.configureGradleEnterprise(gradleEnterprise);
-                overrides.configureBuildCache(buildCache);
-            };
+    private static void applySettingsPlugin(Object gradleEnterpriseExtension, ProviderFactory providers, Settings settings) {
```
runningcode, line 54: @facewindu your indentation and spacing settings are different than that of this repository. can you make them consistent in order to slim up the diff?
it's hard to see that there are only changes to a few lines.

src/main/java/com/gradle/CommonCustomUserDataGradlePlugin.java
```diff
@@ -43,38 +43,47 @@ public void apply(Object target) {
...
+        customGradleEnterpriseConfig.configureBuildScanPublishing(buildScan);
+        CustomBuildScanEnhancements buildScanEnhancements = new CustomBuildScanEnhancements(buildScan, providers, settings.getGradle());
+        buildScanEnhancements.apply();
+
+        BuildCacheConfiguration buildCache = settings.getBuildCache();
+        customGradleEnterpriseConfig.configureBuildCache(buildCache);
+
+        // configuration changes applied in this block will override earlier configuration settings,
+        // including those set in the settings.gradle(.kts)
+        Action<Settings> settingsAction = __ -> {
```
runningcode, line 70: can we keep the 3 underscores and spacing from before to minimize the diff?

runningcode (commented):
from a code standpoint it looks good. we should make sure we are using the same whitespace and coding style as the repo.

src/main/java/com/gradle/CommonCustomUserDataGradlePlugin.java
facewindu: yeah, although the 3 underscores are not needed anymore, they don't collide with 2 underscores, as previously.

runningcode: ah, i see, in that case i think its good to change it for consistency.

facewindu: done

src/main/java/com/gradle/CommonCustomUserDataGradlePlugin.java
facewindu: Most of the diff you see comes from the fact that the `settings.getPluginManager().withPlugin()` callback has been extracted out. Therefore there is one less indentation level.

runningcode (approved):
thanks!

ldaley (approved):

PR #86, opened 2022-08-19
Adding jvm kotlin arguments

cdsap:
Adding Kotlin jvm arguments as tags


runningcode:
Thanks Inaki. We like to keep this repository as things that are appropriate and common for all customers. We also like to keep this code base as clean and as simple as possible and maintain parity with the CCUD Maven extension.

Let's make a github gist for now to make it as easy as possible for customers to use this. If we feel it has broader use, we should consider adding it to [our build config samples](https://github.com/gradle/gradle-enterprise-build-config-samples/tree/main/build-data-capturing-gradle-samples).

PR #88, opened 2022-09-01
Use buildId to calculate TeamCity build URL

britter:


alextu:
I believe this test should be fixed accordingly: https://github.com/gradle/ge/blob/main/common-custom-user-data-gradle-plugin-test-func/src/test/groovy/com/gradle/custom/CommonCustomUserDataGradlePluginFuncTest.groovy#L146

clayburn (approved):
LGTM

etiennestuder:
We should triple-check this PR and with different versions of TC. We keep changing this logic and it always breaks some scenario.

clayburn:
> We should triple-check this PR and with different versions of TC. We keep changing this logic and it always breaks some scenario.
> 
> And we should not forget about the Maven CCUD extension.

👍🏻 This has been tested against Classic and Experimental UI on 2020, 2021, and 2022.

Issue #90, opened 2022-09-27
IDE sync from IntelliJ IDEA incorrectly tagged as Android Studio


erichaagdev:
Build Scans are incorrectly tagged as coming from `Android Studio` when running an IDE sync from IntelliJ IDEA. 

See the following build scan run from IntelliJ IDEA as an example: https://scans.gradle.com/s/eh22ik7skmtge

I verified running a task from IntelliJ IDEA is tagged correctly: https://scans.gradle.com/s/5clegna2kwvfi

Additional information:
- Using `com.gradle.common-custom-user-data-gradle-plugin` version 1.8.1
- Using `com.gradle.enterprise` version 3.11.1
- Using IntelliJ IDEA 2022.2.2 (Ultimate Edition)

runningcode:
https://github.com/gradle/common-custom-user-data-gradle-plugin/pull/91

PR #91, opened 2022-10-04
Fix Intellij tag on IDE syncs

jprinet:
Favor `idea.vendor.name` system property over `android.injected.invoked.from.ide` which is populated during _Intellij_ and _Android studio_ syncs

fixes https://github.com/gradle/common-custom-user-data-gradle-plugin/issues/90

Signed-off-by: Jerome Prinet <jprinet@gradle.com>

### DIFF
--- a/src/main/java/com/gradle/CustomBuildScanEnhancements.java
+++ b/src/main/java/com/gradle/CustomBuildScanEnhancements.java
@@ -10,6 +10,7 @@ import org.gradle.api.tasks.TaskCollection;
 import org.gradle.api.tasks.testing.Test;
 import org.gradle.util.GradleVersion;
 
+import java.util.EnumSet;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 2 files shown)


src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -57,22 +58,57 @@ private void captureOs() {
...
                 Optional<String> androidStudioVersion = projectProperty("android.injected.studio.version");
-                Optional<String> ideaVersion = sysProperty("idea.version");
                 Optional<String> eclipseVersion = sysProperty("eclipse.buildId");
                 Optional<String> ideaSync = sysProperty("idea.sync.active");
                 if (ideaSync.isPresent()) {
                     buildScan.tag("IDE sync");
                 }
-                if (invokedFromAndroidStudio.isPresent()) {
+                if (ideaVendorName.isPresent()) {
+                    IdeaFlavour.getByVendor(ideaVendorName.get()).ifPresent(ideaFlavour -> {
```
runningcode, line 94: can we make this simpler by comparing the strings instead of using an enum?

```
if (ideaVendorName.equalsIgnoreCase("Google")) {
  // tag with Android studio
  else (ideaVendorName.equalsIgnoreCase("Jetbrains") {
  // tag with Intellij
  }
  ```

**Review by runningcode — COMMENTED** [review]
thanks! is there a way to make this simpler? I made a suggestion.

Do we have to make this change in the CCUD maven extension as well?

**Comment by erichaagdev** [comment]
I pulled down these changes to verify they work as expected. To test, I ran an IDE sync from IntelliJ IDEA and generated a build scan before applying the changes and after applying the changes. 

Here is a build scan comparison showing the results: https://ge.solutions-team.gradle.com/c/as4r4nrwsymx2/p4k7hfdlmngyq/task-inputs

**Review by erichaagdev — APPROVED** [review]
Good stuff! Thanks!

**Comment by jprinet** [comment]
> thanks! is there a way to make this simpler? I made a suggestion.

This is surely possible even if I'm not so found of string comparisons

> Do we have to make this change in the CCUD maven extension as well?

Maven extension does not handle Android Studio tag, I don't think you can use something else than Gradle in there
https://github.com/gradle/common-custom-user-data-maven-extension/blob/main/src/main/java/com/gradle/CustomBuildScanEnhancements.java#L54

**Comment by jprinet** [comment]
> I pulled down these changes to verify they work as expected. To test, I ran an IDE sync from IntelliJ IDEA and generated a build scan before and after applying the changes.
> 
> Here is a build scan comparison showing the results: https://ge.solutions-team.gradle.com/c/as4r4nrwsymx2/p4k7hfdlmngyq/task-inputs

Thanks Eric 💪

File: src/main/java/com/gradle/CustomBuildScanEnhancements.java (hunk 1 above)
Comment (jprinet) [reply src/main/java/com/gradle/CustomBuildScanEnhancements.java:94]: Let me adjust the current implementation that way

**Comment by jprinet** [comment]
fixes https://github.com/gradle/common-custom-user-data-gradle-plugin/issues/90

File: src/main/java/com/gradle/CustomBuildScanEnhancements.java (hunk 1 above)
Comment (jprinet) [reply src/main/java/com/gradle/CustomBuildScanEnhancements.java]: You can now review the adjusted implementation @runningcode

File: src/main/java/com/gradle/CustomBuildScanEnhancements.java [hunk 2]
```diff
@@ -61,30 +61,45 @@ private void captureIde() {
         if (!isCi()) {
             // Wait for projects to load to ensure Gradle project properties are initialized
             gradle.projectsEvaluated(g -> {
+                Optional<String> ideaVendorName = sysProperty("idea.vendor.name");
```
alextu, line 64: Cool that there seems to be a relevant property now ! Do you know if it's a recent addition ? I asked a similar question a year ago to Jetbrains and they could not give me such property: https://youtrack.jetbrains.com/issue/IDEA-279187#focus=Comments-27-5233519.0-0

jprinet: This is the information I got

> The idea.vendor.name is available approximately since 2020.1 release

alextu: Nice 👍 !

src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -61,30 +61,45 @@ private void captureIde() {
...
                 } else {
                     buildScan.tag("Cmd Line");
                 }
             });
         }
     }
 
+    private void tagIde(String ideLabel, String version) {
+        buildScan.tag(ideLabel);
+        if(!version.isEmpty()) {
```
runningcode, line 98: i think if we did the null check here we could remove the `orElse("")` and simplify all the callsites. WDYT?

runningcode (commented):
one more comment to simplify things :)

src/main/java/com/gradle/CustomBuildScanEnhancements.java
jprinet: Do you mean passing an optional as method parameter (calling `get()` on a nulled Optional throws an exception)?
this is [discouraged](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Optional.html)

furthermore, doing that would test the optional parameter twice for Eclipse case and this [IntelliJ](https://github.com/gradle/common-custom-user-data-gradle-plugin/pull/91/files#diff-c2a2e523f8b7d89d2a35757da6898ae13ccc09fc7f735c03f820dc06f7956896R84) one

runningcode: I mean like this:
```suggestion
        if(version == null && !version.isEmpty()) {
```

jprinet: `version` wouldn't be null, a `NoSuchElementException` [would be raised](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Optional.html#get())

erichaagdev:
> fixes #90

@jprinet this needs to be in the PR description if you mean to automatically close #90 when this is merged.

etiennestuder:
@runningcode what keeps this PR from being merged?

runningcode (approved):
looks good to me!

PR #95, opened 2022-10-26
Add a link to all build scans with same Jenkins build number and job name

clayburn:
Builds on Jenkins commonly run multiple invocations of Gradle, generating multiple build scans. A common question when looking at a build scan is "which other builds ran in this job/pipeline?". Our current custom links do not easily link back to this query in GE.

Currently, there are 4 custom links for a Jenkins build. One of these links back to Jenkins (based on `BUILD_URL`). The other three link back to custom value searches as follows:

- All CI builds run on a given node (based on `NODE_NAME`)
- All CI builds with a particular job name (based on `JOB_NAME`)
- All CI builds with a particular stage name (based on `STAGE_NAME`)
 
We also capture as a custom value the `BUILD_NUMBER` environment variable as `CI build number`. This CI build number correlates to one run of the pipeline, but the numbers are not unique across all pipelines. So, for example, two distinct pipelines may use `1` as the build number for their first build. So a filter only on `BUILD_NUMBER` may show builds from 2 separate pipelines.

This PR adds a custom link to the list of build scans filtered by _both_ `JOB_NAME` and `BUILD_NUMBER` called "CI build build scans" so that a user can easily see the other builds scans from a particular build in Jenkins.

See https://ge.solutions-team.gradle.com/s/4yg2e7i6ea6ua for an example.

### DIFF
--- a/src/main/java/com/gradle/CustomBuildScanEnhancements.java
+++ b/src/main/java/com/gradle/CustomBuildScanEnhancements.java
@@ -10,6 +10,7 @@ import org.gradle.api.tasks.TaskCollection;
 import org.gradle.api.tasks.testing.Test;
 import org.gradle.util.GradleVersion;
 
+import java.util.AbstractMap;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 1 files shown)


etiennestuder (approved):
LGTM.

@clayburn and I went over it verbally, and decided to change the name of the custom link. Other than that, all good.

etiennestuder:
Once merged, @clayburn will then also do the Maven side.

We decided _not_ to instantly release this feature in new releases but aggregate some more changes first.

etiennestuder (approved):
Latest changes LGTM.

PR #116, opened 2022-12-28
Refactor CI environment variable checks.

runningcode:
The goal of this is to reduce existing code duplication. This will make it easier in the future to refactor all the CI environment variable checks to be done inside a static buildFinished context so that they are configuration cache compatible.
If the approach is okay, we will also need to do the same refactor for the CCUD Maven extension.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ out/
 build/
 target/
 *.DS_Store
+bin/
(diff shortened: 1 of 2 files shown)


alextu:
I understand it's an intermediate state where there's actually a bit more code: `isXXX()` + `static isXXX()` but eventually there should be only `static isXXX()` calls, right ?

erichaagdev (approved):
Approved, but I think adding `bin` to the `.gitignore` can be removed.

runningcode:
I think I will continue working on this PR before reopening. Thanks for the feedback!

PR #117, opened 2023-01-02
Fix createReleaseTag ordering

alextu:
This should ensure `createReleaseTag` is executed after `publishPlugins` whenever `publishPlugins githubRelease` is requested.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -85,12 +85,14 @@ githubRelease {
    body = layout.projectDirectory.file('release/changes.md').asFile.text.trim()
 }
 
-tasks.register('createReleaseTag', CreateGitTag) {
+def createReleaseTag = tasks.register('createReleaseTag', CreateGitTag) {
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -85,12 +85,14 @@ githubRelease {
    body = layout.projectDirectory.file('release/changes.md').asFile.text.trim()
 }
 
-tasks.register('createReleaseTag', CreateGitTag) {
+def createReleaseTag = tasks.register('createReleaseTag', CreateGitTag) {
+    // This makes sure the tag is created only after building and publishing the plugin succeeds
```
runningcode, line 89:
```suggestion
    // Ensure the tag is created iff building and publishing the plugin succeeds.
```

runningcode (approved):
Looks good, I recommended a minor adjustment to the text of the comment. If you agree could you also make sure that text is applied in the other places?

PR #120, opened 2023-01-19
Move captureCiMetadata to execution phase

### LINKED ISSUE #80: Tracking env vars and properties makes configuration-cache not reusable
This plugin reads a lot of env vars, which is great to provide more info, but doing that Gradle won't reuse configuration-cache because there is always some value changing.

Could we change it in a way they are not tracked?

jprinet:
The PR aims to capture CI metadata at execution time (in a `buildFinished` block) not to invalidate configuration cache on CI metadata change (ie. `build number`)

Teamcity data are now fetched from a teamcity generated build file because project properties set in an init file can't easily be read at execution time through a [Provider](https://docs.gradle.org/current/javadoc/org/gradle/api/provider/ProviderFactory.html#gradleProperty-java.lang.String-) due to [this issue](https://github.com/gradle/gradle/issues/23572).

The `projectEvaluated` callback has been replaced with a `afterProject` callback. The latter is called even in case of a configuration cache entry restored.

fixes #80 

Signed-off-by: Jerome Prinet <jprinet@gradle.com>

### DIFF
--- a/src/main/java/com/gradle/CustomBuildScanEnhancements.java
+++ b/src/main/java/com/gradle/CustomBuildScanEnhancements.java
@@ -4,6 +4,7 @@ import com.gradle.scan.plugin.BuildResult;
 import com.gradle.scan.plugin.BuildScanExtension;
 import org.gradle.api.Action;
 import org.gradle.api.Task;
+import org.gradle.api.file.Directory;
 import org.gradle.api.invocation.Gradle;
 import org.gradle.api.provider.ProviderFactory;
 import org.gradle.api.tasks.TaskCollection;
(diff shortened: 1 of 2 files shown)


src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -104,214 +109,196 @@ private void tagIde(String ideLabel, String version) {
...
-                customValueSearchLinker.addCustomValueAndSearchLink("CI stage", value));
-
-            jobName.ifPresent(j -> buildNumber.ifPresent(b -> {
-                Map<String, String> params = new LinkedHashMap<>();
-                params.put("CI job", j);
-                params.put("CI build number", b);
-                customValueSearchLinker.registerLink("CI pipeline", params);
-            }));
-        }
+        gradle.afterProject(project -> {
```
runningcode, line 116: is this `afterProject` still needed given the entire captureCiMetadata is performed inside `buildFinished` ?

src/main/java/com/gradle/CiUtils.java
```diff
@@ -0,0 +1,71 @@
+package com.gradle;
+
+import org.gradle.api.provider.ProviderFactory;
+
+public class CiUtils {
```
runningcode, line 5: Two things,

I think we should make thing change separately and apply it to both the Maven extension and Gradle plugin for easier maintenance.


What do you think of a change where this class is instantiated to avoid passing a `ProviderFactory` to each method?

runningcode (commented):
this is a good start, i think we should split up the PR as proposed in one of my comments.

src/main/java/com/gradle/CustomBuildScanEnhancements.java
jprinet: Yes, if not, you're getting `root project is not yet available for build` when passing the the project directory parameter to the `Action`

src/main/java/com/gradle/CiUtils.java
jprinet: We could do that, let me split the PR

jprinet: This PR is now addressing the issue only

runningcode: Ah sorry, I meant to start with a PR that refactors the checks for CI environment variables. This PR still contains some refactoring for CI environment variable checks. This way we can review the logic changes independently of the refactoring changes and we can apply that to the GE maven extension as well.

jprinet:
Closing in favor of #121 and #123

PR #121, opened 2023-01-24
Extract ci logic to a utility class

jprinet:
CI related methods are use in several places and copy / pasting ([here](https://github.com/gradle/common-custom-user-data-gradle-plugin/blob/97e48144a0d4370d3e2bb409bcba02878adb4aa7/src/main/java/com/gradle/CustomBuildScanEnhancements.java#L273) and [there](https://github.com/gradle/common-custom-user-data-gradle-plugin/blob/97e48144a0d4370d3e2bb409bcba02878adb4aa7/src/main/java/com/gradle/CustomBuildScanEnhancements.java#L404)) is not ideal

This PR aims to centralize this in a utility class, the methods are exposed as static methods. 

Q: Instance Vs static?
Creating an instance of a utility class doesn't simplify the logic. If we do so, we should apply the same pattern to the existing `Utils` class as well.

### DIFF
--- a/src/main/java/com/gradle/CiUtils.java
+++ b/src/main/java/com/gradle/CiUtils.java
@@ -4,6 +4,9 @@ import org.gradle.api.provider.ProviderFactory;
 
 final class CiUtils {
 
+    private CiUtils() {
+    }
+
     static boolean isCi(ProviderFactory providers) {
         return isGenericCI(providers)
                 || isJenkins(providers)
--- a/src/main/java/com/gradle/CiUtils.java
+++ b/src/main/java/com/gradle/CiUtils.java
@@ -1,0 +1,71 @@
+package com.gradle;
+
+import org.gradle.api.provider.ProviderFactory;
+
+final class CiUtils {
+
+    static boolean isCi(ProviderFactory providers) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/gradle/CiUtils.java
```diff
@@ -0,0 +1,74 @@
+package com.gradle;
+
+import org.gradle.api.provider.ProviderFactory;
+
+final class CiUtils {
```
runningcode, line 5: What are your thoughts on naming this just `CI` or `CIs`? I personally dislike the `Utils` naming convention as it ends up making it such that every class is named `XXXUtils`.

runningcode (commented):
Looks great! just one comment

src/main/java/com/gradle/CiUtils.java
jprinet: No strong opinion, I followed the existing pattern 😅 
I find `CiUtils` more meaningful than `CI / CIs` though.

I usually only use the `Utils / Helper` suffix for those classes exposing static methods.

runningcode (approved):
Looks good, please merge and submit the same PR for the CCUD maven extension.

PR #123, opened 2023-01-25
Move CI and IDE metadata processing at execution time

jprinet:
The PR aims to capture CI and IDE metadata at execution time (in a `buildFinished` block), therefore preventing configuration cache invalidation if a property changes (ie. `CI build number`)

Teamcity data are now fetched from a teamcity generated build file as project properties set in an init file can't easily be read at execution time through a [Provider](https://docs.gradle.org/current/javadoc/org/gradle/api/provider/ProviderFactory.html#gradleProperty-java.lang.String-) due to https://github.com/gradle/gradle/issues/23572.

The `projectEvaluated` callback has been replaced by `Providers` used at execution time in order to fix configuration cache restore scenarii.

fixes https://github.com/gradle/common-custom-user-data-gradle-plugin/issues/80

This PR was tested with 
- Gradle 6.0, 7.2, 7.6
- w/o configuration cache
- Jenkins 2.332.3
- Teamcity 2022.10.1 (build 116934)
- Android Studio 2021.3.1 Patch 1

Signed-off-by: Jerome Prinet [jprinet@gradle.com](mailto:jprinet@gradle.com)

### DIFF
--- a/src/main/java/com/gradle/CustomBuildScanEnhancements.java
+++ b/src/main/java/com/gradle/CustomBuildScanEnhancements.java
@@ -22,6 +22,18 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Stream;
 
+import static com.gradle.CiUtils.isAzurePipelines;
+import static com.gradle.CiUtils.isBamboo;
+import static com.gradle.CiUtils.isBitrise;
+import static com.gradle.CiUtils.isCi;
+import static com.gradle.CiUtils.isCircleCI;
+import static com.gradle.CiUtils.isGitHubActions;
+import static com.gradle.CiUtils.isGitLab;
+import static com.gradle.CiUtils.isGoCD;
+import static com.gradle.CiUtils.isHudson;
+import static com.gradle.CiUtils.isJenkins;
+import static com.gradle.CiUtils.isTeamCity;
+import static com.gradle.CiUtils.isTravis;
 import static com.gradle.Utils.appendIfMissing;
 import static com.gradle.Utils.envVariable;
 import static com.gradle.Utils.execAndCheckSuccess;
@@ -75,7 +87,7 @@ final class CustomBuildScanEnhancements {
     }
 
     private void captureIde() {
-        if (!CiUtils.isCi(providers)) {
+        if (!isCi(providers)) {
             // Prepare relevant properties for use at execution time
             Map<String, Provider<String>> ideProperties = new HashMap<>();
             ideProperties.put(SYSTEM_PROP_IDEA_VENDOR_NAME, systemPropertyProvider(SYSTEM_PROP_IDEA_VENDOR_NAME, providers));
@@ -162,11 +174,11 @@ final class CustomBuildScanEnhancements {
     }
 
     private void captureCiOrLocal() {
-        buildScan.tag(CiUtils.isCi(providers) ? "CI" : "LOCAL");
+        buildScan.tag(isCi(providers) ? "CI" : "LOCAL");
     }
 
     private void captureCiMetadata() {
-        if (CiUtils.isCi(providers)) {
+        if (isCi(providers)) {
             // Prepare project directory for use at execution time
             Provider<Directory> projectDirectory = providers.provider(() -> gradle.getRootProject().getLayout().getProjectDirectory());
 
@@ -191,7 +203,7 @@ final class CustomBuildScanEnhancements {
 
         @Override
         public void execute(BuildResult buildResult) {
-            if (CiUtils.isJenkins(providers) || CiUtils.isHudson(providers)) {
+            if (isJenkins(providers) || isHudson(providers)) {
                 Optional<String> buildUrl = envVariable("BUILD_URL", providers);
                 Optional<String> buildNumber = envVariable("BUILD_NUMBER", providers);
                 Optional<String> nodeName = envVariable("NODE_NAME", providers);
@@ -199,7 +211,7 @@ final class CustomBuildScanEnhancements {
                 Optional<String> stageName = envVariable("STAGE_NAME", providers);
 
                 buildUrl.ifPresent(url ->
-                        buildScan.link(CiUtils.isJenkins(providers) ? "Jenkins build" : "Hudson build", url));
+                        buildScan.link(isJenkins(providers) ? "Jenkins build" : "Hudson build", url));
                 buildNumber.ifPresent(value ->
                         buildScan.value("CI build number", value));
                 nodeName.ifPresent(value ->
@@ -217,7 +229,7 @@ final class CustomBuildScanEnhancements {
                 }));
             }
 
-            if (CiUtils.isTeamCity(providers)) {
+            if (isTeamCity(providers)) {
                 Optional<String> teamcityBuildPropertiesFile = envVariable("TEAMCITY_BUILD_PROPERTIES_FILE", providers);
                 if (teamcityBuildPropertiesFile.isPresent()) {
                     Properties buildProperties = readPropertiesFile(teamcityBuildPropertiesFile.get(), providers, projectDirectory.get());
@@ -249,7 +261,7 @@ final class CustomBuildScanEnhancements {
                 }
             }
 
-            if (CiUtils.isCircleCI(providers)) {
+            if (isCircleCI(providers)) {
                 envVariable("CIRCLE_BUILD_URL", providers).ifPresent(url ->
                         buildScan.link("CircleCI build", url));
                 envVariable("CIRCLE_BUILD_NUM", providers).ifPresent(value ->
@@ -260,7 +272,7 @@ final class CustomBuildScanEnhancements {
                         customValueSearchLinker.addCustomValueAndSearchLink("CI workflow", value));
             }
 
-            if (CiUtils.isBamboo(providers)) {
+            if (isBamboo(providers)) {
                 envVariable("bamboo_resultsUrl", providers).ifPresent(url ->
                         buildScan.link("Bamboo build", url));
                 envVariable("bamboo_buildNumber", providers).ifPresent(value ->
@@ -273,7 +285,7 @@ final class CustomBuildScanEnhancements {
                         customValueSearchLinker.addCustomValueAndSearchLink("CI agent", value));
             }
 
-            if (CiUtils.isGitHubActions(providers)) {
+            if (isGitHubActions(providers)) {
                 Optional<String> gitHubRepository = envVariable("GITHUB_REPOSITORY", providers);
                 Optional<String> gitHubRunId = envVariable("GITHUB_RUN_ID", providers);
                 if (gitHubRepository.isPresent() && gitHubRunId.isPresent()) {
@@ -285,7 +297,7 @@ final class CustomBuildScanEnhancements {
                         customValueSearchLinker.addCustomValueAndSearchLink("CI run", value));
             }
 
-            if (CiUtils.isGitLab(providers)) {
+            if (isGitLab(providers)) {
                 envVariable("CI_JOB_URL", providers).ifPresent(url ->
                         buildScan.link("GitLab build", url));
                 envVariable("CI_PIPELINE_URL", providers).ifPresent(url ->
@@ -296,7 +308,7 @@ final class CustomBuildScanEnhancements {
                         customValueSearchLinker.addCustomValueAndSearchLink("CI stage", value));
             }
 
-            if (CiUtils.isTravis(providers)) {
+            if (isTravis(providers)) {
                 envVariable("TRAVIS_BUILD_WEB_URL", providers).ifPresent(url ->
                         buildScan.link("Travis build", url));
                 envVariable("TRAVIS_BUILD_NUMBER", providers).ifPresent(value ->
@@ -306,14 +318,14 @@ final class CustomBuildScanEnhancements {
                 envVariable("TRAVIS_EVENT_TYPE", providers).ifPresent(buildScan::tag);
             }
 
-            if (CiUtils.isBitrise(providers)) {
+            if (isBitrise(providers)) {
                 envVariable("BITRISE_BUILD_URL", providers).ifPresent(url ->
                         buildScan.link("Bitrise build", url));
                 envVariable("BITRISE_BUILD_NUMBER", providers).ifPresent(value ->
                         buildScan.value("CI build number", value));
             }
 
-            if (CiUtils.isGoCD(providers)) {
+            if (isGoCD(providers)) {
                 Optional<String> pipelineName = envVariable("GO_PIPELINE_NAME", providers);
                 Optional<String> pipelineNumber = envVariable("GO_PIPELINE_COUNTER", providers);
                 Optional<String> stageName = envVariable("GO_STAGE_NAME", providers);
@@ -337,7 +349,7 @@ final class CustomBuildScanEnhancements {
                         customValueSearchLinker.addCustomValueAndSearchLink("CI stage", value));
             }
 
-            if (CiUtils.isAzurePipelines(providers)) {
+            if (isAzurePipelines(providers)) {
                 Optional<String> azureServerUrl = envVariable("SYSTEM_TEAMFOUNDATIONCOLLECTIONURI", providers);
                 Optional<String> azureProject = envVariable("SYSTEM_TEAMPROJECT", providers);
                 Optional<String> buildId = envVariable("BUILD_BUILDID", providers);
@@ -426,17 +438,17 @@ final class CustomBuildScanEnhancements {
         }
 
         private String getGitBranchName(Supplier<String> gitCommand) {
-            if (CiUtils.isJenkins(providers) || CiUtils.isHudson(providers)) {
+            if (isJenkins(providers) || isHudson(providers)) {
                 Optional<String> branch = envVariable("BRANCH_NAME", providers);
                 if (branch.isPresent()) {
                     return branch.get();
                 }
-            } else if (CiUtils.isGitLab(providers)) {
+            } else if (isGitLab(providers)) {
                 Optional<String> branch = envVariable("CI_COMMIT_REF_NAME", providers);
                 if (branch.isPresent()) {
                     return branch.get();
                 }
-            } else if (CiUtils.isAzurePipelines(providers)) {
+            } else if (isAzurePipelines(providers)) {
                 Optional<String> branch = envVariable("BUILD_SOURCEBRANCH", providers);
                 if (branch.isPresent()) {
                     return branch.get();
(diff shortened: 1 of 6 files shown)


src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -55,215 +64,289 @@ void apply() {
...
-
-            jobName.ifPresent(j -> buildNumber.ifPresent(b -> {
-                Map<String, String> params = new LinkedHashMap<>();
-                params.put("CI job", j);
-                params.put("CI build number", b);
-                customValueSearchLinker.registerLink("CI pipeline", params);
-            }));
+        if (CiUtils.isCi(providers)) {
+            // Prepare project directory for use at execution time
+            Provider<Directory> projectDirectory = providers.provider(() -> gradle.getRootProject().getLayout().getProjectDirectory());
```
runningcode, line 155: is this Gradle 4 or 5 compatible?

src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -55,215 +64,289 @@ void apply() {
...
+                }
+
+                buildId.ifPresent(value ->
+                        buildScan.value("CI build number", value));
             }
+        }
 
-            buildId.ifPresent(value ->
-                buildScan.value("CI build number", value));
+        private static Properties readPropertiesFile(String fileName, ProviderFactory providers, Directory projectDirectory) {
```
runningcode, line 342: what does this method declaration do? it seems like just a redirect

runningcode (commented):
I know it isn't ready for review but I added two comments. When the PR is ready, please add what versions of Gradle and TeamCity this was tested against :)

src/main/java/com/gradle/CustomBuildScanEnhancements.java
jprinet: Absolutely, this was already there before the PR
https://github.com/gradle/common-custom-user-data-gradle-plugin/blob/97e48144a0d4370d3e2bb409bcba02878adb4aa7/src/main/java/com/gradle/CustomBuildScanEnhancements.java#L527

jprinet: Just removed the indirection, that's better now 👍

src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -60,210 +69,280 @@ private void captureOs() {
...
-            Utils.envVariable("CI_JOB_URL", providers).ifPresent(url ->
-                buildScan.link("GitLab build", url));
-            Utils.envVariable("CI_PIPELINE_URL", providers).ifPresent(url ->
-                buildScan.link("GitLab pipeline", url));
-            Utils.envVariable("CI_JOB_NAME", providers).ifPresent(value ->
-                customValueSearchLinker.addCustomValueAndSearchLink("CI job", value));
-            Utils.envVariable("CI_JOB_STAGE", providers).ifPresent(value ->
-                customValueSearchLinker.addCustomValueAndSearchLink("CI stage", value));
-        }
+            if(CiUtils.isTeamCity(providers)) {
```
runningcode, line 204:
```suggestion
            if (CiUtils.isTeamCity(providers)) {
```
let's fix the formatting from the previous PR

runningcode (commented):
another round of comments. looking good. i just wanted to leave these comments but I haven't finished the review yet

src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -60,210 +69,280 @@ private void captureOs() {
...
+                    if (Utils.isNotEmpty(teamCityBuildNumber)) {
+                        buildScan.value("CI build number", teamCityBuildNumber);
+                    }
+                    String teamCityBuildTypeId = buildProperties.getProperty("teamcity.buildType.id");
+                    if (Utils.isNotEmpty(teamCityBuildTypeId)) {
+                        buildScan.value("CI build config", teamCityBuildTypeId);
+                    }
+                    String teamCityAgentName = buildProperties.getProperty("agent.name");
+                    if (Utils.isNotEmpty(teamCityAgentName)) {
+                        buildScan.value("CI agent", teamCityAgentName);
```
runningcode, line 231: before we were calling the `customValueSearchLinker` why did we change it here? Is it not needed?

runningcode (commented):
Can you also test this with Gradle 4 and 5?
Since TC is not yet CC compatible, did you manually test the TC configuration some other way to make sure the configuration cache was re-usable?
Ok, done for this round of reviews.

src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -60,210 +69,280 @@ private void captureOs() {
...
 
-                if (ideaSync.isPresent()) {
-                    buildScan.tag("IDE sync");
-                }
-            });
+    private Provider<String> systemPropertyProvider(String name, ProviderFactory providers) {
+        if(Utils.isGradle61OrNewer()) {
+            return providers.systemProperty(name);
+        } else {
+            return providers.provider(() -> System.getProperty(name));
```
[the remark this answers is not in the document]
jprinet: The condition is not about deprecation but availability of the method. 
[systemProperty](https://docs.gradle.org/current/javadoc/org/gradle/api/provider/ProviderFactory.html#systemProperty-java.lang.String-) was introduced in 6.1

src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -60,210 +69,280 @@ private void captureOs() {
...
-                } else if (ideaVersion.isPresent()) {
-                    // this case should be handled by the ideaVendorName condition but keeping it for compatibility reason (ideaVendorName started with 2020.1)
-                    tagIde("IntelliJ IDEA", ideaVersion.get());
-                } else if (eclipseVersion.isPresent()) {
-                    tagIde("Eclipse", eclipseVersion.get());
-                } else {
-                    buildScan.tag("Cmd Line");
-                }
+            // Prepare relevant properties for use at execution time
+            Map<String, Provider<String>> relevantProperties = new HashMap<>();
```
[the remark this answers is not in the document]
jprinet: Credit goes to @bamboo there

src/main/java/com/gradle/CustomBuildScanEnhancements.java
jprinet: Well spotted, this one just slipped through the cracks while refactoring, thanks!

src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -60,210 +69,280 @@ private void captureOs() {
...
+            buildScan.buildFinished(new CaptureIdeMetadataAction(buildScan, relevantProperties));
+        }
+    }
 
-                if (ideaSync.isPresent()) {
-                    buildScan.tag("IDE sync");
-                }
-            });
+    private Provider<String> systemPropertyProvider(String name, ProviderFactory providers) {
+        if(Utils.isGradle61OrNewer()) {
```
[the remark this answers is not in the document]
jprinet: I reformatted the code and stop ignored whitespaces in my commit view

src/main/java/com/gradle/CustomBuildScanEnhancements.java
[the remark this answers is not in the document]
runningcode: Is there any reason we cannot use the following code for all Gradle versions?
```
providers.provider(() -> System.getProperty(name))
```

src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -60,210 +69,289 @@ private void captureOs() {
...
         }
     }
 
-    private void captureCiOrLocal() {
-        buildScan.tag(CiUtils.isCi(providers) ? "CI" : "LOCAL");
+    private Provider<String> gradlePropertyProvider(String name, ProviderFactory providers) {
+        if (Utils.isGradle62OrNewer()) {
+            return providers.gradleProperty(name);
+        } else {
+            return providers.provider(() -> (String) gradle.getRootProject().findProperty(name));
```
runningcode, line 98: is there any reason we cannot the following for all Gradle versions?
`providers.provider(() -> (String) gradle.getRootProject().findProperty(name))`

src/main/java/com/gradle/CustomBuildScanEnhancements.java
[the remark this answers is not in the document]
jprinet: We surely could, I prefer to have the idiomatic way clearly highlighted with a condition on Gradle version to be able to get rid of legacy way once we decide not to support Gradle < 6.1.

src/main/java/com/gradle/CustomBuildScanEnhancements.java
jprinet: same reason as `providers.systemProperty`

jprinet:
> Can you also test this with Gradle 4 and 5? Since TC is not yet CC compatible, did you manually test the TC configuration some other way to make sure the configuration cache was re-usable? Ok, done for this round of reviews.

I tested with Gradle 4.1 and 5.
It was a good call as [getOrElse](https://docs.gradle.org/current/javadoc/org/gradle/api/provider/Provider.html#getOrElse-T-) was introduced in 4.3, I added a condition to enable 4.1 compatibility.

Regarding Teamcity use case, the only data read is an environment variable at execution time. Jenkins case is processed the same way and was successfully tested with configuration cache enabled.

src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -55,215 +71,294 @@ void apply() {
...
+            }
+
+            if (CiUtils.isTeamCity(providers)) {
+                Optional<String> teamcityBuildPropertiesFile = envVariable("TEAMCITY_BUILD_PROPERTIES_FILE", providers);
+                if (teamcityBuildPropertiesFile.isPresent()) {
+                    Properties buildProperties = readPropertiesFile(teamcityBuildPropertiesFile.get(), providers, projectDirectory.get());
+
+                    String teamcityConfigFile = buildProperties.getProperty("teamcity.configuration.properties.file");
+                    if (isNotEmpty(teamcityConfigFile)) {
+                        Properties configProperties = readPropertiesFile(teamcityConfigFile, providers, projectDirectory.get());
```
runningcode, line 227: In the previous version we would only read the file if the build id was also present. why did we change this? I'm not saying either way is wrong or correct, I just want to understand.

src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -55,215 +71,294 @@ void apply() {
...
-                } else {
-                    buildScan.tag("Cmd Line");
-                }
-
-                if (ideaSync.isPresent()) {
-                    buildScan.tag("IDE sync");
-                }
-            });
+            // Prepare relevant properties for use at execution time
+            Map<String, Provider<String>> relevantProperties = new HashMap<>();
```
runningcode, line 80: can we name this variable `ideProperties`
```suggestion
            Map<String, Provider<String>> ideProperties = new HashMap<>();
```

src/main/java/com/gradle/CustomBuildScanEnhancements.java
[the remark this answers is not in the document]
runningcode: I see. This seems like an interesting trade off of best practices vs code simplicity. Let's keep it but can we move this to the bottom of the file in a static method where the other methods are? (along with the `gradlePropertyProvider`)

runningcode (commented):
Ok, I've added my review. :)
Looking good. I like the static import changes. I think we should be consistent about their uses though. Let's either always use them or never use them. In this case I think we should always use them.
Can we double check that on TeamCity (and spot check some other CI systems) that all the links in the build scan appear the same as in the previous released version of CCUD? Please also add a build scan link in the PR with the before and after.

src/main/java/com/gradle/CustomBuildScanEnhancements.java
jprinet: I found it easier to read and assumption is that build id is always present.
Let's revert to the original approach though

jprinet:
> Ok, I've added my review. :) Looking good. I like the static import changes. I think we should be consistent about their uses though. Let's either always use them or never use them. In this case I think we should always use them. Can we double check that on TeamCity (and spot check some other CI systems) that all the links in the build scan appear the same as in the previous released version of CCUD? Please also add a build scan link in the PR with the before and after.

I aligned the static imports as suggested 👍

Thanks as well for asking to check the navigation links, I realized that those were not displayed due to an overlooked [execution in buildFinished block](https://github.com/gradle/common-custom-user-data-gradle-plugin/blob/a5589b79b04ee9fff6dc2a519409c7b60aa64a9f/src/main/java/com/gradle/CustomBuildScanEnhancements.java#L372).
I adjusted the code to now process the links in sequence ([related commit](https://github.com/gradle/common-custom-user-data-gradle-plugin/pull/123/commits/49cfb0d730caa4253c36529df9b24b05d05d93c3)). I also decided to have `captureCiMetadata` and `captureGitMetadata` to each have its own instance of the `CustomValueSearchLinker` to simplify process ordering.

src/main/java/com/gradle/CustomBuildScanEnhancements.java
runningcode: I didn't mean to suggest anything, I was just curious if there was a reason behind it. :)

src/main/java/com/gradle/CustomBuildScanEnhancements.java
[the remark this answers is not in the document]
runningcode: sounds good.

src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -353,63 +464,56 @@ private String getGitBranchName(Supplier<String> gitCommand) {
     }
 
     /**
-     * Collects custom values that should have a search link, and creates these links in `buildFinished`.
-     * The actual construction of the links must be deferred to ensure the Server URL is set.
+     * Collects all custom values that should have a search link and finally creates the links to ensure deduplication.
```
runningcode, line 467: Which ones are duplicated here? And how does the de-duplication work exactly if there are multiple instances of this class?

I think the main reason that this existed was to defer the construction of the links until the server url is set. Can we be sure that all uses of this class have the server URL set already? If so do we still need this class?

runningcode (commented):
Getting close here! One thing, could you please add the before and after build scan links, I would also expect the outcome to be the same and that is why the links are useful.

src/main/java/com/gradle/CustomBuildScanEnhancements.java
jprinet: Re-reading the implementation there is indeed no duplication, post-processing the navigation link is therefore useless.
I thought there were some dynamic links for a given CI server with `key` extracted from the environment but this is just not the case.
The class is still useful though to encapsulate the logic imo.

There is still one case where the server URL could not be set as `captureGitMetadata` is occurring in a `background` block which could run at configuration time depending on Gradle version.
Let me fix this.

jprinet:
Build scans without changes:
[Jenkins with CC stored](https://ge.solutions-team.gradle.com/s/tiwk3bu7ch4mk)
[Jenkins with CC not restored](https://ge.solutions-team.gradle.com/s/r7ujkayqyjnwq)
[Teamcity](https://ge.solutions-team.gradle.com/s/yliychbb45f4q)

Build scans with changes:
[Jenkins with CC stored](https://ge.solutions-team.gradle.com/s/b5b7h66hju5nq)
[Jenkins with CC restored](https://ge.solutions-team.gradle.com/s/qzm7fgb7o2yfe)
[Teamcity](https://ge.solutions-team.gradle.com/s/cke6ynofnh7o6)

src/main/java/com/gradle/CustomBuildScanEnhancements.java
jprinet: I pushed a new commit which remove `CustomValueSearchLinker` and fixes the _server URL_ retrieval inside `captureGitMetadata`

jprinet:
Confirming that I was able to get CC restored in Teamcity with a custom init script.

src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -352,75 +456,6 @@ private String getGitBranchName(Supplier<String> gitCommand) {
...
-            // the parameters for a link querying multiple custom values look like:
-            // search.names=name1,name2&search.values=value1,value2
-            // this reduction groups all names and all values together in order to properly generate the query
-            values.entrySet().stream()
-                .sorted(Map.Entry.comparingByKey()) // results in a deterministic order of link parameters
-                .reduce((a, b) -> new AbstractMap.SimpleEntry<>(a.getKey() + "," + b.getKey(), a.getValue() + "," + b.getValue()))
-                .ifPresent(x -> registerLink(linkLabel, x.getKey(), x.getValue()));
-        }
-
-        private synchronized void registerLink(String linkLabel, String name, String value) {
```
runningcode, line 396: Do you know why we needed the synchronized here before? Is it safe to remove?

etiennestuder: It's accessing a non-thread safe map in `customValueLinks.put` (could be invoked from multiple bg jobs in parallel).

jprinet: This is not required as there is no state to keep anymore. The navigation links are added inline.

src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -352,75 +456,6 @@ private String getGitBranchName(Supplier<String> gitCommand) {
...
-                return buildScan.getServer();
-            } catch (NoSuchMethodException e) {
-                // not available in Gradle 4.x / Build Scan plugin 1.x
-                return null;
-            }
-        }
-
-    }
-
     private void captureTestParallelization() {
```
runningcode, line 459: can we move this to the bottom of the file like it used to be to lessen the diff? or is there a reason this grouping makes more logical sense?

runningcode (commented):
Thanks just one comment and I think this is good to merge! Thanks for all the effort and checking everything here.

src/main/java/com/gradle/CustomBuildScanEnhancements.java
jprinet: I grouped all the static methods together at the end of the file, there was originally only 1, we could indeed move the `captureTestParallelization`  method but I find the actual setup better.

runningcode (approved):
Great work! Let's think about a release plan before merging.

jprinet:
Duplicate of #129, Duplicate of #130, Duplicate of #131

PR #130, opened 2023-02-20
Capture IDE related metadata at execution time

jprinet:
The PR aims to capture CI metadata and IDE metadata at execution time (in a `buildFinished` block), therefore preventing configuration cache invalidation if a property changes (ie. `CI build number`) and allow correct execution of configuration cache restore scenarii.

Teamcity metadata are now fetched from a build file generated by Teamcity, as project properties set via an init script can't be easily read at execution time through a [Provider](https://docs.gradle.org/current/javadoc/org/gradle/api/provider/ProviderFactory.html#gradleProperty-java.lang.String-) due to https://github.com/gradle/gradle/issues/23572.

fixes https://github.com/gradle/common-custom-user-data-gradle-plugin/issues/80

This PR was tested with
- Gradle 6.0, 7.2, 7.6, 8.0.1
- w/o configuration cache
- Jenkins 2.332.3
- Teamcity 2022.10.1 (build 116934)

Signed-off-by: Jerome Prinet [jprinet@gradle.com](mailto:jprinet@gradle.com)

### DIFF
--- a/release/changes.md
+++ b/release/changes.md
@@ -1,2 +1,3 @@
 - [NEW] Add a link to all build scans with same Jenkins build number and job name #95
-- [FIX] - Support overrides to GradleEnterpriseBuildCache connector
+- [FIX] Support overrides to GradleEnterpriseBuildCache connector
+- [FIX] Move CI and IDE metadata processing at execution time (https://github.com/gradle/common-custom-user-data-gradle-plugin/issues/80) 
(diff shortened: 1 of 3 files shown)


runningcode:
Was this (and the following PR) tested with Gradle 4? I remember you also tested it with Gradle 4 at some point.

Now that Gradle 8 is out, can we also test it with Gradle 8 and the `STABLE_CONFIGURATION_CACHE` flag? It adds some stricter checks. I want to make sure we are always compatible.

jprinet:
> Was this (and the following PR) tested with Gradle 4? I remember you also tested it with Gradle 4 at some point.
> 
> Now that Gradle 8 is out, can we also test it with Gradle 8 and the `STABLE_CONFIGURATION_CACHE` flag? It adds some stricter checks. I want to make sure we are compatible when the strictest checks are enabled.

I'll check what's up with Gradle 8 and 4, can't recall if this was tested with the latter

PR #136, opened 2023-03-02
Reorder methods to have Maven vs Gradle CCUD diff more consistent

jprinet:


runningcode:
This looks good. I think we decided to put all static methods at the bottom. Can we make these static just so we follow that rule?

runningcode (approved):

PR #142, opened 2023-03-15
User can configure BT and GE build cache connector with the same sys prop `gradle.cache.remote.url`

etiennestuder:
- Using the GE build cache connector, the remote build cache URL provided via `gradle.cache.remote.url` is split up and the _server_ and _path_ parts set individually on the GE connector API
- The behavior of configuring the BT and the GE connector is now more similar in terms replacing/appending a path provided via `gradle.cache.remote.path`

### DIFF
--- a/src/main/java/com/gradle/Overrides.java
+++ b/src/main/java/com/gradle/Overrides.java
@@ -13,7 +13,7 @@ import java.time.Duration;
 import java.util.Optional;
 
 import static com.gradle.Utils.appendIfMissing;
-import static com.gradle.Utils.prependIfMissing;
+import static com.gradle.Utils.prependAndAppendIfMissing;
 import static com.gradle.Utils.stripPrefix;
 
 /**
@@ -106,7 +106,7 @@ final class Overrides {
 
     private static URI replacePath(URI uri, String path) {
         try {
-            String finalPath = appendIfMissing(prependIfMissing("/", path), "/");
+            String finalPath = prependAndAppendIfMissing(path, "/");
             return new URI(uri.getScheme(), uri.getUserInfo(), uri.getHost(), uri.getPort(), finalPath, uri.getQuery(), uri.getFragment());
         } catch (URISyntaxException e) {
             throw new IllegalArgumentException("Cannot construct URI: " + uri, e);
@@ -115,7 +115,7 @@ final class Overrides {
 
     private static URI appendPath(URI uri, String path) {
         try {
-            String currentPath = appendIfMissing(prependIfMissing("/", uri.getPath()), "/");
+            String currentPath = prependAndAppendIfMissing(uri.getPath(), "/");
             String additionalPath = appendIfMissing(stripPrefix("/", path), "/");
             String finalPath = currentPath + additionalPath;
             return new URI(uri.getScheme(), uri.getUserInfo(), uri.getHost(), uri.getPort(), finalPath, uri.getQuery(), uri.getFragment());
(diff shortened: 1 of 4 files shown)


src/main/java/com/gradle/Overrides.java
```diff
@@ -99,4 +104,44 @@ private static String toEnvVarName(String sysPropertyName) {
         return sysPropertyName.toUpperCase().replace('.', '_');
     }
 
+    private static URI replacePath(URI uri, String path) {
+        try {
+            String finalPath = appendIfMissing(prependIfMissing("/", path), "/");
```
clayburn, line 109: I think this append(prepend("/", path), "/") pattern could be made a little clearer if refactored as one function call.

src/main/java/com/gradle/Utils.java
```diff
@@ -82,29 +82,12 @@ static String stripPrefix(String prefix, String string) {
...
-    }
-
-    static URI appendPathAndTrailingSlash(URI baseUri, String path) {
-        if (isNotEmpty(path)) {
-            String normalizedBasePath = appendIfMissing(baseUri.getPath(), "/");
-            String normalizedPath = appendIfMissing(stripPrefix("/", path), "/");
-            return baseUri.resolve(normalizedBasePath).resolve(normalizedPath);
-        }
-        return baseUri;
+    static String prependIfMissing(String suffix, String str) {
```
clayburn, line 85: The `suffix` parameter should be renamed to `prefix`.

src/main/java/com/gradle/Utils.java
```diff
@@ -82,29 +82,12 @@ static String stripPrefix(String prefix, String string) {
...
-        if (isNotEmpty(basePath)) {
-            if (isNotEmpty(path)) {
-                String normalizedBasePath = appendIfMissing(basePath, "/");
-                String normalizedPath = stripPrefix("/", path);
-                return normalizedBasePath + normalizedPath;
-            }
-            return basePath;
-        }
-        return path;
+    static String appendIfMissing(String str, String suffix) {
```
clayburn, line 89: I find it surprising that `appendIfMissing` and `prependIfMissing` swap the order of parameters. I think the parameters are trying to match the order of the resulting string, but for two methods that are doing very similar things, I would expect the method signatures to match.

clayburn (approved):
Functionally this looks good to me. I made a few refactoring suggestions.

src/main/java/com/gradle/Utils.java
etiennestuder: I thought about this, but I find the calling code easier to read the current way.

clayburn: That's fine. Just an opinion in this case 🙂

src/main/java/com/gradle/Utils.java
etiennestuder: +1

src/main/java/com/gradle/Overrides.java
etiennestuder: So, a `prependAndAppendIfMissing`? +1

clayburn: That name sounds good to me.

etiennestuder: Done.

src/main/java/com/gradle/Utils.java
etiennestuder: Done.

src/main/java/com/gradle/Utils.java
etiennestuder: I've applied your suggestion. Looks indeed better.

PR #143, opened 2023-03-15
User can configure to allow insecure protocol on remote build cache connector

etiennestuder:


clayburn (approved):
LGTM. See also comments on #142.

src/main/java/com/gradle/Overrides.java
```diff
@@ -52,6 +56,7 @@ void configureGradleEnterpriseOnGradle4(BuildScanExtension buildScan) {
         booleanSysPropertyOrEnvVariable(GRADLE_ENTERPRISE_ALLOW_UNTRUSTED_SERVER, providers).ifPresent(buildScan::setAllowUntrustedServer);
     }
 
+    @SuppressWarnings("DataFlowIssue")
```
erichaagdev, line 59: Why is this suppressed? What issues with "data flow" might there be?

README.md
```diff
@@ -88,25 +88,27 @@ to modify the build scripts. For example, to disable the local build cache when
...
+|:--------------------------------------------|:---------------------------------------------|:-------------------------------------------|
+| buildCache.remote.setEnabled                | gradle.cache.remote.enabled                  | GRADLE_CACHE_REMOTE_ENABLED                |
+| buildCache.remote.setPush                   | gradle.cache.remote.push                     | GRADLE_CACHE_REMOTE_PUSH                   |
+| buildCache.remote.setAllowUntrustedServer   | gradle.cache.remote.allowUntrustedServer     | GRADLE_CACHE_REMOTE_ALLOWUNTRUSTEDSERVER   |
+| buildCache.remote.setAllowInsecureProtocol  | gradle.cache.remote.setAllowInsecureProtocol | GRADLE_CACHE_REMOTE_ALLOWINSECUREPROTOCOL  |
+| buildCache.remote.setServer/setPath         | gradle.cache.remote.url                      | GRADLE_CACHE_REMOTE_URL                    |
+| buildCache.remote.setPath                   | gradle.cache.remote.path                     | GRADLE_CACHE_REMOTE_PATH                   |
+| buildCache.remote.setPath                   | gradle.cache.remote.shard                    | GRADLE_CACHE_REMOTE_SHARD                  |
```
erichaagdev, line 111: It seems there are now competing properties, but the precedence isn't explained here.

src/main/java/com/gradle/Overrides.java
etiennestuder: A null check for the remote URL.

README.md
etiennestuder: This is not new. It typically makes no sense to override more than one at a time.

PR #145, opened 2023-03-15
Fixes to issues in URL Overrides for the GradleEnterpriseBuildCacheConnector

clayburn:
With the latest set of changes, two scenarios fail:

- Defining a URL without a path component
- Defining a URL without a path component + a shard

With the latest set of changes, defining a URL override for the GradleEnterpriseBuildCacheConnecter without a path fails to properly connect to the build cache. For example, a build that specifies `-Dgradle.cache.remote.url=https://ge.solutions-team.gradle.com` will set the path as an empty string and result in errors like these: https://ge.solutions-team.gradle.com/s/5rtouxcb2nwdk/console-log?page=1#L14

This PR fixes this by avoiding a call to setPath if the derived path is essentially empty (i.e. '/' or an empty string).

Once that is fixed, defining a URL override without a path and a shard together throws a NullPointerException. This is address by assuming that the path is "cache" when `getPath` returns null.

### DIFF
--- a/src/main/java/com/gradle/Overrides.java
+++ b/src/main/java/com/gradle/Overrides.java
@@ -80,7 +80,7 @@ final class Overrides {
         } else if (buildCache.getRemote() instanceof GradleEnterpriseBuildCache) {
             buildCache.remote(GradleEnterpriseBuildCache.class, remote -> {
                 sysPropertyOrEnvVariable(REMOTE_CACHE_URL, providers).map(Overrides::serverOnly).ifPresent(remote::setServer);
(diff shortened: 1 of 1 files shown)


etiennestuder:
`For example, a build that specifies -Dgradle.cache.remote.url=https://ge.solutions-team.gradle.com will set the path as an empty string and result in errors like these: https://ge.solutions-team.gradle.com/s/5rtouxcb2nwdk/console-log?page=1#L14`

This is a bug in the GE plugin and has been fixed on main yesterday.

etiennestuder:
Thanks for the investigation and fixes, @clayburn. With my PR #146, these fixes are no longer needed because the functionality is gone.

PR #150, opened 2023-03-20
Support build scan links to alternate GitHub URLs

### LINKED ISSUE #109: Fix Build Scan links when running on GitHub Enterprise Server
It is possible to use the GitHub actions with GitHub Enterprise Server. This means that any reference to `github.com` in Build Scan links will be incorrect. For example: 

https://github.com/gradle/common-custom-user-data-gradle-plugin/blob/3908b28b83348c4485bf5d760690b852f4256dba/src/main/java/com/gradle/CustomBuildScanEnhancements.java#L187

Instead of hardcoding `github.com` we should key off of the `GITHUB_SERVER_URL` environment variable. According to the documentation, this environment variable is available on [Enterprise Server](https://docs.github.com/en/enterprise-server@3.7/actions/learn-github-actions/environment-variables#default-environment-variables), [Enterprise Cloud](https://docs.github.com/en/enterprise-cloud@latest/actions/learn-github-actions/environment-variables#default-environment-variables), and [Free](https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables).

This aligns with the GitHub documentation as well:

> If you need to use a workflow run's URL from within a job, you can combine these environment variables: `$GITHUB_SERVER_URL/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID`

tylerbertrand:
Fixes #109

When creating links to GitHub Actions builds, the plugin now references the environment variable `GITHUB_SERVER_URL` instead of hardcoding `https://github.com/`.

When creating build scan links to the remote repository, the plugin now uses a regex to parse out the correct url instead of hardcoding `https://github.com/` or `https://gitlab.com`.

### Local Tests

Published the CCUD plugin to Maven local for testing.

HTTP URL variations tested:
* `https://github.com/gradle/common-custom-user-data-gradle-plugin.git`: https://ge.solutions-team.gradle.com/s/r4phrtqudyt2y
* `https://github.com:443/gradle/common-custom-user-data-gradle-plugin.git`: https://ge.solutions-team.gradle.com/s/lbibg7tgoq6ea
* `https://github.somecompany.com/gradle/common-custom-user-data-gradle-plugin.git`: https://ge.solutions-team.gradle.com/s/njk2dgwc74j6y
* `https://github.somecompany.com//gradle/common-custom-user-data-gradle-plugin.git`: https://ge.solutions-team.gradle.com/s/zz6ohiobmry7c
* `https://gh.somecompany.com/gradle/common-custom-user-data-gradle-plugin.git` (no link is generated): https://ge.solutions-team.gradle.com/s/qdxgbegbcfrt6

SSH URL variations tested:
* `git@github.com:gradle/common-custom-user-data-gradle-plugin.git`: https://ge.solutions-team.gradle.com/s/wjuxfu2vcgewc
* `ssh://git@github.com/gradle/common-custom-user-data-gradle-plugin.git`: https://ge.solutions-team.gradle.com/s/f34u7kk3y4csc
* `ssh://git@github.com:22/gradle/common-custom-user-data-gradle-plugin.git`: https://ge.solutions-team.gradle.com/s/uvk6gngtx4nxo
* `git@github.somecompany.com:gradle/common-custom-user-data-gradle-plugin.git`: https://ge.solutions-team.gradle.com/s/axzabanengtm6
* `git@gh.somecompany.com:gradle/common-custom-user-data-gradle-plugin.git` (no link is generated): https://ge.solutions-team.gradle.com/s/32oxx7u7vw4oi

### CI Tests

Published the CCUD plugin `1.10.1` to the plugin portal for CI testing.

A branch was created `tylerbertrand/test-pr-150` and pushed that uses `1.10.1` from the staging portal to test the custom link generation in CI builds.

[This GitHub Actions build](https://github.com/gradle/common-custom-user-data-gradle-plugin/actions/runs/4633501276) created the build scan https://ge.solutions-team.gradle.com/s/3e3cjt4ncw2oo. The correct `GitHub source` custom link was created, and based on the tags, it is clear the build executed in CI. It can be verified [here](https://ge.solutions-team.gradle.com/s/3e3cjt4ncw2oo/build-dependencies?focusedDependency=WzAsMSwxLFswLDEsWzFdXV0&toggled=W1swXSxbMCwxXV0) that the CCUD plugin used was `1.10.1` from the staging portal.

### DIFF
--- a/release/changes.md
+++ b/release/changes.md
@@ -1,1 +1,1 @@
-[FIX] Generating incorrect "GitHub Actions Build" and "GitHub/GitLab Source" links to repositories not hosted at github.com or gitlab.com
+[FIX] Generates incorrect "GitHub Actions Build" and "GitHub/GitLab Source" links to repositories not hosted at github.com or gitlab.com
--- a/release/changes.md
+++ b/release/changes.md
@@ -1,0 +1,1 @@
+[FIX] Support build scan links to alternate GitHub URLs (such as GitHub Enterprise servers)
--- a/release/changes.md
+++ b/release/changes.md
@@ -1,1 +1,1 @@
-[FIX] Support build scan links to alternate GitHub URLs (such as GitHub Enterprise servers)
+[FIX] Generating incorrect "GitHub Actions Build" and "GitHub/GitLab Source" links to repositories not hosted at github.com or gitlab.com
--- a/src/main/java/com/gradle/CustomBuildScanEnhancements.java
+++ b/src/main/java/com/gradle/CustomBuildScanEnhancements.java
@@ -440,16 +440,24 @@ final class CustomBuildScanEnhancements {
 
         private Optional<URI> toWebRepoUri(String gitRepoUri) {
             Matcher matcher = Pattern.compile("^(?:https://|(?:ssh)?.*?@)(.*?(?:github|gitlab).*?)(?:/|:[0-9]*?/|:)(.*?)(?:\\.git)?$").matcher(gitRepoUri);
-            if(!matcher.matches()) {
+            if (matcher.matches()) {
+                String scheme = "https";
+                String host = matcher.group(1);
+                String path = matcher.group(2).startsWith("/") ? matcher.group(2) : "/" + matcher.group(2);
+                return toUri(scheme, host, path);
+            } else {
                 return Optional.empty();
             }
+        }
+
+        private Optional<URI> toUri(String scheme, String host, String path) {
             try {
-                String path = matcher.group(2).startsWith("/") ? matcher.group(2) : "/" + matcher.group(2);
-                return Optional.of(new URI("https", matcher.group(1), path, null));
+                return Optional.of(new URI(scheme, host, path, null));
             } catch (URISyntaxException e) {
                 return Optional.empty();
             }
         }
+
     }
 
     private void captureTestParallelization() {
@@ -530,4 +538,5 @@ final class CustomBuildScanEnhancements {
             return providers.provider(() -> (String) gradle.getRootProject().findProperty(name));
         }
     }
+
 }
--- a/src/main/java/com/gradle/CustomBuildScanEnhancements.java
+++ b/src/main/java/com/gradle/CustomBuildScanEnhancements.java
@@ -401,15 +401,12 @@ final class CustomBuildScanEnhancements {
             if (gitHubUrl.isPresent() && gitRepository.isPresent() && isNotEmpty(gitCommitId)) {
                 buildScan.link("GitHub source", gitHubUrl.get() + "/" + gitRepository.get() + "/tree/" + gitCommitId);
             } else if (isNotEmpty(gitRepo) && isNotEmpty(gitCommitId)) {
-                Optional<URI> gitRepoUri = toUri(gitRepo);
-                gitRepoUri.ifPresent(r -> {
-                    String gitRepoPath = r.getPath().endsWith(".git") ? r.getPath().substring(0, r.getPath().length() - 4) : r.getPath();
+                Optional<URI> webRepoUri = toWebRepoUri(gitRepo);
+                webRepoUri.ifPresent(r -> {
                     if (r.getHost().contains("github")) {
-                        toHttpsUri(r.getHost(), gitRepoPath + "/tree/" + gitCommitId)
-                            .ifPresent(u ->  buildScan.link("GitHub source", u.toString()));
+                        buildScan.link("GitHub source", r + "/tree/" + gitCommitId);
                     } else if (r.getHost().contains("gitlab")) {
-                        toHttpsUri(r.getHost(), gitRepoPath + "/-/commit/" + gitCommitId)
-                            .ifPresent(u -> buildScan.link("GitLab source", u.toString()));
+                        buildScan.link("GitLab source", r + "/-/commit/" + gitCommitId);
                     }
                 });
             }
@@ -439,17 +436,21 @@ final class CustomBuildScanEnhancements {
             return gitCommand.get();
         }
 
-        private Optional<URI> toUri(String uri) {
+        private Optional<URI> toWebRepoUri(String gitRepoUri) {
             try {
-                return Optional.of(new URI(uri));
+                URI tempUri = new URI(gitRepoUri);
+                String gitRepoPath = tempUri.getPath().endsWith(".git") ? tempUri.getPath().substring(0, tempUri.getPath().length() - 4) : tempUri.getPath();
+                return Optional.of(new URI("https", tempUri.getHost(), gitRepoPath, null));
             } catch (URISyntaxException e) {
                 return Optional.empty();
             }
         }
 
-        private Optional<URI> toHttpsUri(String host, String path) {
+        private Optional<URI> stripDotGitExtension(URI uri) {
+            String path = uri.getPath();
             try {
-                return Optional.of(new URI("https", host, path));
+                return path.endsWith(".git") ? Optional.of(new URI(uri.getScheme(), uri.getHost(), path.substring(0, path.length() - 4)))
+                    : Optional.of(uri);
             } catch (URISyntaxException e) {
                 return Optional.empty();
             }
--- a/src/main/java/com/gradle/CustomBuildScanEnhancements.java
+++ b/src/main/java/com/gradle/CustomBuildScanEnhancements.java
@@ -397,19 +397,15 @@ final class CustomBuildScanEnhancements {
             }
 
             if (isNotEmpty(gitRepo) && isNotEmpty(gitCommitId)) {
-                if (gitRepo.contains("github.com/") || gitRepo.contains("github.com:")) {
-                    Matcher matcher = Pattern.compile("(.*)github\\.com[/|:](.*)").matcher(gitRepo);
-                    if (matcher.matches()) {
-                        String rawRepoPath = matcher.group(2);
-                        String repoPath = rawRepoPath.endsWith(".git") ? rawRepoPath.substring(0, rawRepoPath.length() - 4) : rawRepoPath;
-                        buildScan.link("Github source", "https://github.com/" + repoPath + "/tree/" + gitCommitId);
-                    }
-                } else if (gitRepo.contains("gitlab.com/") || gitRepo.contains("gitlab.com:")) {
-                    Matcher matcher = Pattern.compile("(.*)gitlab\\.com[/|:](.*)").matcher(gitRepo);
-                    if (matcher.matches()) {
-                        String rawRepoPath = matcher.group(2);
-                        String repoPath = rawRepoPath.endsWith(".git") ? rawRepoPath.substring(0, rawRepoPath.length() - 4) : rawRepoPath;
-                        buildScan.link("GitLab Source", "https://gitlab.com/" + repoPath + "/-/commit/" + gitCommitId);
+                Matcher matcher = Pattern.compile("(https://|git@)(.+?)[:|/](.*)").matcher(gitRepo);
+                if (matcher.matches()) {
+                    String repoUrl = "https://" + matcher.group(2) + "/";
+                    String rawRepoPath = matcher.group(3);
+                    String repoPath = rawRepoPath.endsWith(".git") ? rawRepoPath.substring(0, rawRepoPath.length() - 4) : rawRepoPath;
+                    if (gitRepo.contains("github")) {
+                        buildScan.link("Github source", repoUrl + repoPath + "/tree/" + gitCommitId);
+                    } else if (gitRepo.contains("gitlab")) {
+                        buildScan.link("GitLab Source", repoUrl + repoPath + "/-/commit/" + gitCommitId);
                     }
                 }
             }
--- a/src/main/java/com/gradle/CustomBuildScanEnhancements.java
+++ b/src/main/java/com/gradle/CustomBuildScanEnhancements.java
@@ -438,7 +438,7 @@ final class CustomBuildScanEnhancements {
 
         private Optional<URI> toWebRepoUri(String gitRepoUri) {
             try {
-                URI tempUri = new URI(gitRepoUri);
+                URI tempUri = new URI(preprocessSshRepoUri(gitRepoUri));
                 String gitRepoPath = tempUri.getPath().endsWith(".git") ? tempUri.getPath().substring(0, tempUri.getPath().length() - 4) : tempUri.getPath();
                 return Optional.of(new URI("https", tempUri.getHost(), gitRepoPath, null));
             } catch (URISyntaxException e) {
@@ -446,14 +446,15 @@ final class CustomBuildScanEnhancements {
             }
         }
 
-        private Optional<URI> stripDotGitExtension(URI uri) {
-            String path = uri.getPath();
-            try {
-                return path.endsWith(".git") ? Optional.of(new URI(uri.getScheme(), uri.getHost(), path.substring(0, path.length() - 4)))
-                    : Optional.of(uri);
-            } catch (URISyntaxException e) {
-                return Optional.empty();
+        private String preprocessSshRepoUri(String gitRepoUri) {
+            if (!gitRepoUri.startsWith("ssh://")
+                && gitRepoUri.contains("@")
+                && gitRepoUri.indexOf(':', gitRepoUri.indexOf('@')) != -1) {
+                StringBuilder sshGitRepoUri = new StringBuilder("ssh://").append(gitRepoUri);
+                sshGitRepoUri.setCharAt(sshGitRepoUri.indexOf(":", sshGitRepoUri.indexOf("@")), '/');
+                return sshGitRepoUri.toString();
             }
+            return gitRepoUri;
         }
     }
 
@@ -535,5 +536,4 @@ final class CustomBuildScanEnhancements {
             return providers.provider(() -> (String) gradle.getRootProject().findProperty(name));
         }
     }
-
 }
--- a/src/main/java/com/gradle/CustomBuildScanEnhancements.java
+++ b/src/main/java/com/gradle/CustomBuildScanEnhancements.java
@@ -269,10 +269,10 @@ final class CustomBuildScanEnhancements {
 
             if (isGitHubActions(providers)) {
                 Optional<String> gitHubUrl = envVariable("GITHUB_SERVER_URL", providers);
-                Optional<String> gitHubRepository = envVariable("GITHUB_REPOSITORY", providers);
+                Optional<String> gitRepository = envVariable("GITHUB_REPOSITORY", providers);
                 Optional<String> gitHubRunId = envVariable("GITHUB_RUN_ID", providers);
-                if (gitHubUrl.isPresent() && gitHubRepository.isPresent() && gitHubRunId.isPresent()) {
-                    buildScan.link("GitHub Actions build", gitHubUrl + "/" + gitHubRepository.get() + "/actions/runs/" + gitHubRunId.get());
+                if (gitHubUrl.isPresent() && gitRepository.isPresent() && gitHubRunId.isPresent()) {
+                    buildScan.link("GitHub Actions build", gitHubUrl + "/" + gitRepository.get() + "/actions/runs/" + gitHubRunId.get());
                 }
                 envVariable("GITHUB_WORKFLOW", providers).ifPresent(value ->
                         addCustomValueAndSearchLink(buildScan, "CI workflow", value));
@@ -397,9 +397,9 @@ final class CustomBuildScanEnhancements {
             }
 
             Optional<String> gitHubUrl = envVariable("GITHUB_SERVER_URL", providers);
-            Optional<String> gitHubRepository = envVariable("GITHUB_REPOSITORY", providers);
-            if (gitHubUrl.isPresent() && gitHubRepository.isPresent() && isNotEmpty(gitCommitId)) {
-                buildScan.link("GitHub source", gitHubUrl.get() + "/" + gitHubRepository.get() + "/tree/" + gitCommitId);
+            Optional<String> gitRepository = envVariable("GITHUB_REPOSITORY", providers);
+            if (gitHubUrl.isPresent() && gitRepository.isPresent() && isNotEmpty(gitCommitId)) {
+                buildScan.link("GitHub source", gitHubUrl.get() + "/" + gitRepository.get() + "/tree/" + gitCommitId);
             } else if (isNotEmpty(gitRepo) && isNotEmpty(gitCommitId)) {
                 Optional<URI> gitRepoUri = toUri(gitRepo);
                 gitRepoUri.ifPresent(r -> {
--- a/src/main/java/com/gradle/CustomBuildScanEnhancements.java
+++ b/src/main/java/com/gradle/CustomBuildScanEnhancements.java
@@ -402,10 +402,10 @@ final class CustomBuildScanEnhancements {
                 if (gitHubUrl.isPresent() && gitHubRepository.isPresent()) {
                     buildScan.link("GitHub source", gitHubUrl + "/" + gitHubRepository + "/tree/" + gitCommitId);
                 } else {
-                    Matcher matcher = Pattern.compile("(https://|git@)(.*?(github|gitlab).*?)[:|/](.*)").matcher(gitRepo);
+                    Matcher matcher = Pattern.compile("((https|ssh)://|.*?@)(.*?(github|gitlab).*?)[:/](.*)").matcher(gitRepo);
                     if (matcher.matches()) {
-                        String repoUrl = "https://" + matcher.group(2) + "/";
-                        String rawRepoPath = matcher.group(4);
+                        String repoUrl = "https://" + matcher.group(3) + "/";
+                        String rawRepoPath = matcher.group(5);
                         String repoPath = rawRepoPath.endsWith(".git") ? rawRepoPath.substring(0, rawRepoPath.length() - 4) : rawRepoPath;
                         if (gitRepo.contains("github")) {
                             buildScan.link("GitHub source", repoUrl + repoPath + "/tree/" + gitCommitId);
--- a/src/main/java/com/gradle/CustomBuildScanEnhancements.java
+++ b/src/main/java/com/gradle/CustomBuildScanEnhancements.java
@@ -402,10 +402,10 @@ final class CustomBuildScanEnhancements {
                 if (gitHubUrl.isPresent() && gitHubRepository.isPresent()) {
                     buildScan.link("GitHub source", gitHubUrl + "/" + gitHubRepository + "/tree/" + gitCommitId);
                 } else {
-                    Matcher matcher = Pattern.compile("((https|ssh)://|.*?@)(.*?(github|gitlab).*?)[:/](.*)").matcher(gitRepo);
+                    Matcher matcher = Pattern.compile("(https://|.*?@)(.*?(github|gitlab).*?)[:/](.*)").matcher(gitRepo);
                     if (matcher.matches()) {
-                        String repoUrl = "https://" + matcher.group(3) + "/";
-                        String rawRepoPath = matcher.group(5);
+                        String repoUrl = "https://" + matcher.group(2) + "/";
+                        String rawRepoPath = matcher.group(4);
                         String repoPath = rawRepoPath.endsWith(".git") ? rawRepoPath.substring(0, rawRepoPath.length() - 4) : rawRepoPath;
                         if (gitRepo.contains("github")) {
                             buildScan.link("GitHub source", repoUrl + repoPath + "/tree/" + gitCommitId);
--- a/src/main/java/com/gradle/CustomBuildScanEnhancements.java
+++ b/src/main/java/com/gradle/CustomBuildScanEnhancements.java
@@ -361,6 +361,8 @@ final class CustomBuildScanEnhancements {
 
     private static final class CaptureGitMetadataAction implements Action<BuildScanExtension> {
 
+        private static final Pattern GIT_REPO_URI_PATTERN = Pattern.compile("^(?:https://|(?:ssh)?.*?@)(.*?(?:github|gitlab).*?)(?:/|:[0-9]*?/|:)(.*?)(?:\\.git)?$");
+
         private final ProviderFactory providers;
 
         private CaptureGitMetadataAction(ProviderFactory providers) {
@@ -404,11 +406,11 @@ final class CustomBuildScanEnhancements {
                 buildScan.link("GitHub source", gitHubUrl.get() + "/" + gitRepository.get() + "/tree/" + gitCommitId);
             } else if (isNotEmpty(gitRepo) && isNotEmpty(gitCommitId)) {
                 Optional<URI> webRepoUri = toWebRepoUri(gitRepo);
-                webRepoUri.ifPresent(r -> {
-                    if (r.getHost().contains("github")) {
-                        buildScan.link("GitHub source", r + "/tree/" + gitCommitId);
-                    } else if (r.getHost().contains("gitlab")) {
-                        buildScan.link("GitLab source", r + "/-/commit/" + gitCommitId);
+                webRepoUri.ifPresent(uri -> {
+                    if (uri.getHost().contains("github")) {
+                        buildScan.link("GitHub source", uri + "/tree/" + gitCommitId);
+                    } else if (uri.getHost().contains("gitlab")) {
+                        buildScan.link("GitLab source", uri + "/-/commit/" + gitCommitId);
                     }
                 });
             }
@@ -439,7 +441,7 @@ final class CustomBuildScanEnhancements {
         }
 
         private Optional<URI> toWebRepoUri(String gitRepoUri) {
-            Matcher matcher = Pattern.compile("^(?:https://|(?:ssh)?.*?@)(.*?(?:github|gitlab).*?)(?:/|:[0-9]*?/|:)(.*?)(?:\\.git)?$").matcher(gitRepoUri);
+            Matcher matcher = GIT_REPO_URI_PATTERN.matcher(gitRepoUri);
             if (matcher.matches()) {
                 String scheme = "https";
                 String host = matcher.group(1);
--- a/src/main/java/com/gradle/CustomBuildScanEnhancements.java
+++ b/src/main/java/com/gradle/CustomBuildScanEnhancements.java
@@ -268,10 +268,11 @@ final class CustomBuildScanEnhancements {
             }
 
             if (isGitHubActions(providers)) {
+                Optional<String> gitHubUrl = envVariable("GITHUB_SERVER_URL", providers);
                 Optional<String> gitHubRepository = envVariable("GITHUB_REPOSITORY", providers);
                 Optional<String> gitHubRunId = envVariable("GITHUB_RUN_ID", providers);
-                if (gitHubRepository.isPresent() && gitHubRunId.isPresent()) {
-                    buildScan.link("GitHub Actions build", "https://github.com/" + gitHubRepository.get() + "/actions/runs/" + gitHubRunId.get());
+                if (gitHubUrl.isPresent() && gitHubRepository.isPresent() && gitHubRunId.isPresent()) {
+                    buildScan.link("GitHub Actions build", gitHubUrl.get() + "/" + gitHubRepository.get() + "/actions/runs/" + gitHubRunId.get());
                 }
                 envVariable("GITHUB_WORKFLOW", providers).ifPresent(value ->
                         addCustomValueAndSearchLink(buildScan, "CI workflow", value));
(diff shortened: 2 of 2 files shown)


src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -268,10 +268,11 @@ public void execute(BuildResult buildResult) {
             }
 
             if (isGitHubActions(providers)) {
+                Optional<String> gitHubUrl = envVariable("GITHUB_SERVER_URL", providers);
                 Optional<String> gitHubRepository = envVariable("GITHUB_REPOSITORY", providers);
                 Optional<String> gitHubRunId = envVariable("GITHUB_RUN_ID", providers);
-                if (gitHubRepository.isPresent() && gitHubRunId.isPresent()) {
-                    buildScan.link("GitHub Actions build", "https://github.com/" + gitHubRepository.get() + "/actions/runs/" + gitHubRunId.get());
+                if (gitHubUrl.isPresent() && gitHubRepository.isPresent() && gitHubRunId.isPresent()) {
```
jprinet, line 274: Are we sure `GITHUB_SERVER_URL` is always set? I'm wondering if with some old runners or on-prem runners, it could be interesting to fall back to `github.com` if empty

src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -396,19 +397,15 @@ public void execute(BuildScanExtension buildScan) {
...
-                        String repoPath = rawRepoPath.endsWith(".git") ? rawRepoPath.substring(0, rawRepoPath.length() - 4) : rawRepoPath;
-                        buildScan.link("Github source", "https://github.com/" + repoPath + "/tree/" + gitCommitId);
-                    }
-                } else if (gitRepo.contains("gitlab.com/") || gitRepo.contains("gitlab.com:")) {
-                    Matcher matcher = Pattern.compile("(.*)gitlab\\.com[/|:](.*)").matcher(gitRepo);
-                    if (matcher.matches()) {
-                        String rawRepoPath = matcher.group(2);
-                        String repoPath = rawRepoPath.endsWith(".git") ? rawRepoPath.substring(0, rawRepoPath.length() - 4) : rawRepoPath;
-                        buildScan.link("GitLab Source", "https://gitlab.com/" + repoPath + "/-/commit/" + gitCommitId);
+                Matcher matcher = Pattern.compile("(https://|git@)(.+?)[:|/](.*)").matcher(gitRepo);
```
jprinet, line 400: Shouldn't that be limited to `Github` / `Gitlab`?
Not sure if with this approach we could catch other cases as well (_Bitbucket_...)?
Not a big deal as you are verifying afterward, but I would avoid extra computation if not required.

jprinet (changes requested):
Good job, I would just make sure we address the case of `GITHUB_SERVER_URL` unset 👍

release/changes.md
```diff
@@ -0,0 +1 @@
+[FIX] Support build scan links to alternate GitHub URLs (such as GitHub Enterprise servers)
```
etiennestuder, line 1: The fix should describe the problem -- not the solution.

etiennestuder (changes requested):

src/main/java/com/gradle/CustomBuildScanEnhancements.java
tylerbertrand: `GITHUB_SERVER_URL` will only be expected to be set if executing on a GitHub Actions runner though, right?

src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -396,19 +397,15 @@ public void execute(BuildScanExtension buildScan) {
...
-                        String rawRepoPath = matcher.group(2);
-                        String repoPath = rawRepoPath.endsWith(".git") ? rawRepoPath.substring(0, rawRepoPath.length() - 4) : rawRepoPath;
-                        buildScan.link("GitLab Source", "https://gitlab.com/" + repoPath + "/-/commit/" + gitCommitId);
+                Matcher matcher = Pattern.compile("(https?://|git@)(.*?(github|gitlab).*?)[:|/](.*)").matcher(gitRepo);
+                if (matcher.matches()) {
+                    String repoUrl = "https://" + matcher.group(2) + "/";
+                    String rawRepoPath = matcher.group(4);
+                    String repoPath = rawRepoPath.endsWith(".git") ? rawRepoPath.substring(0, rawRepoPath.length() - 4) : rawRepoPath;
+                    if (gitRepo.contains("github")) {
+                        buildScan.link("Github source", repoUrl + repoPath + "/tree/" + gitCommitId);
```
etiennestuder, line 406: `GitHub` instead of `Github`

src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -396,19 +397,15 @@ public void execute(BuildScanExtension buildScan) {
...
+                    } else if (gitRepo.contains("gitlab")) {
+                        buildScan.link("GitLab Source", repoUrl + repoPath + "/-/commit/" + gitCommitId);
```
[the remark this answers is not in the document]
tylerbertrand: Not sure I'm following. `gitHubUrl` is in another nested class `CaptureCiMetadataAction`. Do you mean to just go ahead and reference `GITHUB_SERVER_URL` and get the `repoUrl` from that if it is set?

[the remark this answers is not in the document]
tylerbertrand: Just wanted to follow up on this @etiennestuder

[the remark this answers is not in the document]
etiennestuder: What I'm suggesting is that we also use gitHubUrl when creating the custom links to the source code. Right now, these custom links are only generated for gitlab.com and github.com, if I'm not mistaken. But, when hosting GitHub Enterprise at gh.foo.com, the source code links should also be created.

Or am I misunderstanding something here?

src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -396,19 +397,15 @@ public void execute(BuildScanExtension buildScan) {
...
-                    }
-                } else if (gitRepo.contains("gitlab.com/") || gitRepo.contains("gitlab.com:")) {
-                    Matcher matcher = Pattern.compile("(.*)gitlab\\.com[/|:](.*)").matcher(gitRepo);
-                    if (matcher.matches()) {
-                        String rawRepoPath = matcher.group(2);
-                        String repoPath = rawRepoPath.endsWith(".git") ? rawRepoPath.substring(0, rawRepoPath.length() - 4) : rawRepoPath;
-                        buildScan.link("GitLab Source", "https://gitlab.com/" + repoPath + "/-/commit/" + gitCommitId);
+                Matcher matcher = Pattern.compile("(https?://|git@)(.*?(github|gitlab).*?)[:|/](.*)").matcher(gitRepo);
+                if (matcher.matches()) {
+                    String repoUrl = "https://" + matcher.group(2) + "/";
```
etiennestuder, line 402: The above code is hard-coded to https, but in the pattern matching there will also be a match for http. I would remove http from the pattern that is matched against.

etiennestuder (changes requested):

src/main/java/com/gradle/CustomBuildScanEnhancements.java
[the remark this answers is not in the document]
tylerbertrand: Fixed

src/main/java/com/gradle/CustomBuildScanEnhancements.java
tylerbertrand: Done - if `GITHUB_SERVER_URL` is unset, `https://github.com` is now used as the default value for `gitHubUrl`.

release/changes.md
tylerbertrand: Updated the fix to describe the problem being addressed

src/main/java/com/gradle/CustomBuildScanEnhancements.java
tylerbertrand: Done - the matcher will no longer match on `http://` urls

src/main/java/com/gradle/CustomBuildScanEnhancements.java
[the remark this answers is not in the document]
tylerbertrand: So something like this...
```
Optional<String> gitHubUrl = envVariable("GITHUB_SERVER_URL", providers);
Optional<String> gitHubRepository = envVariable("GITHUB_REPOSITORY", providers);
if (gitHubUrl.isPresent() && gitHubRepository.isPresent()) {
    buildScan.link("GitHub source", gitHubUrl + "/" + gitHubRepository + "/tree/" + gitCommitId);
} else {
    // Generate build scan link to source using pattern matching
}
```
...would get you a build scan link to any enterprise github installation if running in GitHub Actions (or if those environment variables are otherwise set), no matter what that url is. Otherwise, just use the pattern matching already in this PR. Does that track with what you're thinking?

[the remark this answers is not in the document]
etiennestuder: `would get you a build scan link to any enterprise github installation if running in GitHub Actions ` --> yes, that's what I mean.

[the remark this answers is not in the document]
tylerbertrand: Ok cool, sounds good. I'll get that updated.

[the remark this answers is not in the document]
tylerbertrand: Done - the build scan link generation to github source has been updated following the example posted above using `GITHUB_SERVER_URL` and `GITHUB_REPOSITORY`

clayburn (changes requested):

src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -396,19 +397,21 @@ public void execute(BuildScanExtension buildScan) {
...
-                if (gitRepo.contains("github.com/") || gitRepo.contains("github.com:")) {
-                    Matcher matcher = Pattern.compile("(.*)github\\.com[/|:](.*)").matcher(gitRepo);
-                    if (matcher.matches()) {
-                        String rawRepoPath = matcher.group(2);
-                        String repoPath = rawRepoPath.endsWith(".git") ? rawRepoPath.substring(0, rawRepoPath.length() - 4) : rawRepoPath;
-                        buildScan.link("Github source", "https://github.com/" + repoPath + "/tree/" + gitCommitId);
-                    }
-                } else if (gitRepo.contains("gitlab.com/") || gitRepo.contains("gitlab.com:")) {
-                    Matcher matcher = Pattern.compile("(.*)gitlab\\.com[/|:](.*)").matcher(gitRepo);
+                Optional<String> gitHubUrl = envVariable("GITHUB_SERVER_URL", providers);
```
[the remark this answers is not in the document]
tylerbertrand: It's handled below by the line `if (gitHubUrl.isPresent() && gitHubRepository.isPresent()) {`

If either of those is unset, it falls back to the pattern matcher.

src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -396,19 +397,21 @@ public void execute(BuildScanExtension buildScan) {
...
+                Optional<String> gitHubRepository = envVariable("GITHUB_REPOSITORY", providers);
+                if (gitHubUrl.isPresent() && gitHubRepository.isPresent()) {
+                    buildScan.link("GitHub source", gitHubUrl + "/" + gitHubRepository + "/tree/" + gitCommitId);
+                } else {
+                    Matcher matcher = Pattern.compile("(https://|git@)(.*?(github|gitlab).*?)[:|/](.*)").matcher(gitRepo);
```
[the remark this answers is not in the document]
tylerbertrand: `gitHubUrl` will only be populated if running in github actions. If you're running a local build, `GITHUB_SERVER_URL` (and `GITHUB_REPOSITORY`) can't be expected to be set.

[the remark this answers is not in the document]
tylerbertrand: Good catch on the pipe. I'll get that removed.

When would we expect an `ssh://` url? A repository I cloned with ssh reports this:
```bash
% git config --get remote.origin.url
git@github.com:gradle/gradle-enterprise-build-config-samples.git
```
And a repository I cloned with https reports this:
```
% git config --get remote.origin.url
https://github.com/gradle/gradle-enterprise-build-config-samples.git
```

For matching on only the `git` username, I'm admittedly not familiar with when/why that username might change. Though the pattern can definitely be updated to match on any username.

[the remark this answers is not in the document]
tylerbertrand: I pushed a change that matches on `ssh://` or `https://` as well as any ssh username such `git@` or `someuser@`. Let me know what you think.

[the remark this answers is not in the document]
clayburn: > When would we expect an `d` url?

It is a valid way to clone over ssh. I looked a little more, and most of the repositories I've cloned do not have that ssh protocol defined. I think I just happened to find one that did. So I think I would recommend reverting back to the way it was, without the explicit `ssh://` handling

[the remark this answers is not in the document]
clayburn: > If you're running a local build, GITHUB_SERVER_URL (and GITHUB_REPOSITORY) can't be expected to be set.

Good point, I was too focused on CI builds here.

src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -396,19 +397,21 @@ public void execute(BuildScanExtension buildScan) {
...
+                Optional<String> gitHubRepository = envVariable("GITHUB_REPOSITORY", providers);
+                if (gitHubUrl.isPresent() && gitHubRepository.isPresent()) {
+                    buildScan.link("GitHub source", gitHubUrl + "/" + gitHubRepository + "/tree/" + gitCommitId);
+                } else {
+                    Matcher matcher = Pattern.compile("(https://|.*?@)(.*?(github|gitlab).*?)[:/](.*)").matcher(gitRepo);
                     if (matcher.matches()) {
-                        String rawRepoPath = matcher.group(2);
+                        String repoUrl = "https://" + matcher.group(2) + "/";
+                        String rawRepoPath = matcher.group(4);
                         String repoPath = rawRepoPath.endsWith(".git") ? rawRepoPath.substring(0, rawRepoPath.length() - 4) : rawRepoPath;
```
clayburn, line 409: I think we could remove this logic that strips `.git` by adding it as a group to the end of the regular expression above: `...[:/](.*)(?:\.git)`. Then the regular expression would completely capture all of the string manipulation we are doing here.

src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -396,19 +397,21 @@ public void execute(BuildScanExtension buildScan) {
...
-                        buildScan.link("Github source", "https://github.com/" + repoPath + "/tree/" + gitCommitId);
-                    }
-                } else if (gitRepo.contains("gitlab.com/") || gitRepo.contains("gitlab.com:")) {
-                    Matcher matcher = Pattern.compile("(.*)gitlab\\.com[/|:](.*)").matcher(gitRepo);
+                Optional<String> gitHubUrl = envVariable("GITHUB_SERVER_URL", providers);
+                Optional<String> gitHubRepository = envVariable("GITHUB_REPOSITORY", providers);
+                if (gitHubUrl.isPresent() && gitHubRepository.isPresent()) {
+                    buildScan.link("GitHub source", gitHubUrl + "/" + gitHubRepository + "/tree/" + gitCommitId);
+                } else {
+                    Matcher matcher = Pattern.compile("(https://|.*?@)(.*?(github|gitlab).*?)[:/](.*)").matcher(gitRepo);
```
clayburn, line 405: Up to your discretion, but it may make the intent of the regular expression a little more obvious by using non-capturing groups for the groups we aren't actually using.

clayburn (approved):

src/main/java/com/gradle/CustomBuildScanEnhancements.java
tylerbertrand: Agreed, that should clean things up a bit.

Addressed in 4a5f2ca

src/main/java/com/gradle/CustomBuildScanEnhancements.java
tylerbertrand: Good point, definitely the way to go - addressed in 31c5a8f

release/changes.md
```diff
@@ -0,0 +1 @@
+[FIX] Generating incorrect "GitHub Actions Build" and "GitHub/GitLab Source" links to repositories not hosted at github.com or gitlab.com
```
etiennestuder, line 1: Please replace with `Generates`

src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -396,19 +397,20 @@ public void execute(BuildScanExtension buildScan) {
...
-                        String rawRepoPath = matcher.group(2);
-                        String repoPath = rawRepoPath.endsWith(".git") ? rawRepoPath.substring(0, rawRepoPath.length() - 4) : rawRepoPath;
-                        buildScan.link("Github source", "https://github.com/" + repoPath + "/tree/" + gitCommitId);
-                    }
-                } else if (gitRepo.contains("gitlab.com/") || gitRepo.contains("gitlab.com:")) {
-                    Matcher matcher = Pattern.compile("(.*)gitlab\\.com[/|:](.*)").matcher(gitRepo);
+                Optional<String> gitHubUrl = envVariable("GITHUB_SERVER_URL", providers);
+                Optional<String> gitHubRepository = envVariable("GITHUB_REPOSITORY", providers);
+                if (gitHubUrl.isPresent() && gitHubRepository.isPresent()) {
+                    buildScan.link("GitHub source", gitHubUrl + "/" + gitHubRepository + "/tree/" + gitCommitId);
```
etiennestuder, line 403: Did you really test this? It seems to me that you have an implicit conversion from an Optional to a String here, which would not lead to a properly rendered URL.

src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -396,19 +397,20 @@ public void execute(BuildScanExtension buildScan) {
...
-                if (gitRepo.contains("github.com/") || gitRepo.contains("github.com:")) {
-                    Matcher matcher = Pattern.compile("(.*)github\\.com[/|:](.*)").matcher(gitRepo);
-                    if (matcher.matches()) {
-                        String rawRepoPath = matcher.group(2);
-                        String repoPath = rawRepoPath.endsWith(".git") ? rawRepoPath.substring(0, rawRepoPath.length() - 4) : rawRepoPath;
-                        buildScan.link("Github source", "https://github.com/" + repoPath + "/tree/" + gitCommitId);
-                    }
-                } else if (gitRepo.contains("gitlab.com/") || gitRepo.contains("gitlab.com:")) {
-                    Matcher matcher = Pattern.compile("(.*)gitlab\\.com[/|:](.*)").matcher(gitRepo);
+                Optional<String> gitHubUrl = envVariable("GITHUB_SERVER_URL", providers);
```
etiennestuder, line 400: The `isNotEmpty(gitRepo)` check is only needed for the inner `else` block and thus should be moved inwards.

src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -396,19 +397,20 @@ public void execute(BuildScanExtension buildScan) {
...
+                Optional<String> gitHubRepository = envVariable("GITHUB_REPOSITORY", providers);
+                if (gitHubUrl.isPresent() && gitHubRepository.isPresent()) {
+                    buildScan.link("GitHub source", gitHubUrl + "/" + gitHubRepository + "/tree/" + gitCommitId);
+                } else {
+                    Matcher matcher = Pattern.compile("(?:https://|.*?@)(.*?(?:github|gitlab).*?)[:/](.*)\\.git").matcher(gitRepo);
```
etiennestuder, line 405: Why do you do all this pattern matching magic stuff instead of just creating a URI object?

src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -396,19 +397,20 @@ public void execute(BuildScanExtension buildScan) {
...
                     if (matcher.matches()) {
-                        String rawRepoPath = matcher.group(2);
-                        String repoPath = rawRepoPath.endsWith(".git") ? rawRepoPath.substring(0, rawRepoPath.length() - 4) : rawRepoPath;
-                        buildScan.link("GitLab Source", "https://gitlab.com/" + repoPath + "/-/commit/" + gitCommitId);
+                        String repoUrl = "https://" + matcher.group(1) + "/";
+                        String repoPath = matcher.group(2);
+                        if (gitRepo.contains("github")) {
```
etiennestuder, line 409: Why do you need this string check instead of just relying on the check if we are running GitHub actions (`isGitHubActions`) or GitLab (`isGitLab`)?

etiennestuder (changes requested):

src/main/java/com/gradle/CustomBuildScanEnhancements.java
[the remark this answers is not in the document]
tylerbertrand: Here I don't think we want to fall back to `https://github.com` - if this is empty we need to get the url from the output of `git config --get remote.origin.url`.

src/main/java/com/gradle/CustomBuildScanEnhancements.java
tylerbertrand: I'll take a look at those - though this needs to handle local builds as well.

src/main/java/com/gradle/CustomBuildScanEnhancements.java
tylerbertrand: Yeah that was definitely an oversight - I'll get that fixed.

tylerbertrand: Fixed. [Here](https://ge.solutions-team.gradle.com/s/c62popbv5t6d6) is a build scan from a build run in GitHub Actions (verified by the presence of the `CI` tag and `GitHub Actions build` link) that generates a correct `GitHub source` link.

I published the fix as CCUD version `1.10.2` to the staging portal, and we can verify [here](https://ge.solutions-team.gradle.com/s/c62popbv5t6d6/plugins#com.gradle.common-custom-user-data-gradle-plugin-20) that version `1.10.2` was used by that build scan.

src/main/java/com/gradle/CustomBuildScanEnhancements.java
tylerbertrand: Updated to use `URI`. I agree that is the way to go. It is much clearer than the regex and also validates the URL returned by `git config --get remote.origin.url`.

src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -395,21 +400,22 @@ public void execute(BuildScanExtension buildScan) {
...
+                try {
+                    URI gitRepoUri = new URI(gitRepo);
+                    String gitRepoHost = gitRepoUri.getHost();
+                    String gitRepoPath = gitRepoUri.getPath().endsWith(".git") ? gitRepoUri.getPath().substring(0, gitRepoUri.getPath().length() - 4) : gitRepoUri.getPath();
+                    if (gitRepo.contains("github")) {
+                        buildScan.link("GitHub source", "https://" + gitRepoHost + gitRepoPath + "/tree/" + gitCommitId);
+                    } else if (gitRepo.contains("gitlab")) {
+                        buildScan.link("GitLab source", "https://" + gitRepoHost + gitRepoPath + "/-/commit/" + gitCommitId);
                     }
+                } catch (URISyntaxException e) {
```
etiennestuder, line 417: This try-catch is way too wide.

src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -395,21 +400,22 @@ public void execute(BuildScanExtension buildScan) {
...
+                    logger.warn("Remote git repository " + gitRepo + " is not a valid URI");
```
etiennestuder, line 418: Why do you want to log this scenario of the Git repo being an invalid URL?

src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -268,10 +270,11 @@ public void execute(BuildResult buildResult) {
             }
 
             if (isGitHubActions(providers)) {
+                String gitHubUrl = envVariable("GITHUB_SERVER_URL", providers).orElse("https://github.com");
```
etiennestuder, line 273: In which scenarios is the GITHUB_SERVER_URL not present when invoked from a GH action (and the need for a fallback)?

src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -395,21 +400,22 @@ public void execute(BuildScanExtension buildScan) {
...
-                    }
-                } else if (gitRepo.contains("gitlab.com/") || gitRepo.contains("gitlab.com:")) {
-                    Matcher matcher = Pattern.compile("(.*)gitlab\\.com[/|:](.*)").matcher(gitRepo);
-                    if (matcher.matches()) {
-                        String rawRepoPath = matcher.group(2);
-                        String repoPath = rawRepoPath.endsWith(".git") ? rawRepoPath.substring(0, rawRepoPath.length() - 4) : rawRepoPath;
-                        buildScan.link("GitLab Source", "https://gitlab.com/" + repoPath + "/-/commit/" + gitCommitId);
+            Optional<String> gitHubUrl = envVariable("GITHUB_SERVER_URL", providers);
+            Optional<String> gitHubRepository = envVariable("GITHUB_REPOSITORY", providers);
+            if (gitHubUrl.isPresent() && gitHubRepository.isPresent()) {
```
etiennestuder, line 405: Why don't you check for an empty gitCommitId here?

src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -395,21 +400,22 @@ public void execute(BuildScanExtension buildScan) {
...
+                buildScan.link("GitHub source", gitHubUrl.get() + "/" + gitHubRepository.get() + "/tree/" + gitCommitId);
+            } else if (isNotEmpty(gitRepo) && isNotEmpty(gitCommitId)) {
+                try {
+                    URI gitRepoUri = new URI(gitRepo);
+                    String gitRepoHost = gitRepoUri.getHost();
+                    String gitRepoPath = gitRepoUri.getPath().endsWith(".git") ? gitRepoUri.getPath().substring(0, gitRepoUri.getPath().length() - 4) : gitRepoUri.getPath();
+                    if (gitRepo.contains("github")) {
```
etiennestuder, line 412: Under what circumstances would GITHUB_SERVER_URL and GITHUB_REPOSITORY not be present but it's still a repository with the name GitHub in it? A local build?

etiennestuder (changes requested):

src/main/java/com/gradle/CustomBuildScanEnhancements.java
tylerbertrand: So the reason I did it that way is that all the code in the `try` block is reliant on the URI being valid (i.e. the `URISyntaxException` not occurring). At that point, if we don't have a valid URI, we can't execute any of the following code constructing the link, so I felt putting it in the `try` block was appropriate. Is there a better pattern I should have followed for this type of situation?

src/main/java/com/gradle/CustomBuildScanEnhancements.java
tylerbertrand: I didn't really think that was a possible scenario, I was thinking maybe just in the case of a freshly initialized repository without commits? But even then it returns an error message, so I'm not sure when it would be empty. So we could potentially check that it is a valid hash?

src/main/java/com/gradle/CustomBuildScanEnhancements.java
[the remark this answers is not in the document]
tylerbertrand: Yes I think so, I'll get that updated.

tylerbertrand: Correct - for local builds we would not expect those environment variables to be present, but if the remote repository is hosted on a GitHub server, it may have "github" in the hostname.

src/main/java/com/gradle/CustomBuildScanEnhancements.java
tylerbertrand: I could not find any specific scenarios where we would expect them to not be present when invoked from a GitHub Action, but as @jprinet mentioned it *may* be possible given some older runners or on-prem runners. But maybe a fallback isn't the correct way to handle those environment variables missing? Not generating the link may be more correct.

src/main/java/com/gradle/CustomBuildScanEnhancements.java
tylerbertrand: To close the loop on this - those methods in `CiUtils` use CI-specific environment variables to determine the type of environment they're executing in. As this needs to work for local builds as well, I don't think those methods can be used here.

src/main/java/com/gradle/CustomBuildScanEnhancements.java
tylerbertrand: I pushed commit 299d446 that reduces the scope of the try/catch - wdyt?

src/main/java/com/gradle/CustomBuildScanEnhancements.java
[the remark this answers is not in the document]
tylerbertrand: Done

src/main/java/com/gradle/CustomBuildScanEnhancements.java
etiennestuder: CLosing this thread since it has been clarified in the spec.

src/main/java/com/gradle/CustomBuildScanEnhancements.java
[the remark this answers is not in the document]
etiennestuder: Got it. I would not do the fallback in the other place, neither.

src/main/java/com/gradle/CustomBuildScanEnhancements.java
etiennestuder: Got it.

src/main/java/com/gradle/CustomBuildScanEnhancements.java
etiennestuder: I'll propose another change.

src/main/java/com/gradle/CustomBuildScanEnhancements.java
etiennestuder: This seems such an arbitrary scenario that could go wrong that I would not deal with explicitly.

src/main/java/com/gradle/CustomBuildScanEnhancements.java
etiennestuder: I would not add the fallback since you cannot be sure the build runs on github.com, so worst case, the generated links are wrong. IMO, no link is better than a wrong link.

src/main/java/com/gradle/CustomBuildScanEnhancements.java
etiennestuder: You check for an empty commit id in the block below.

src/main/java/com/gradle/CustomBuildScanEnhancements.java
etiennestuder: Got it.

src/main/java/com/gradle/CustomBuildScanEnhancements.java
etiennestuder: Done.

src/main/java/com/gradle/CustomBuildScanEnhancements.java
etiennestuder: Done.

src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -395,22 +396,21 @@ public void execute(BuildScanExtension buildScan) {
...
+            Optional<String> gitRepository = envVariable("GITHUB_REPOSITORY", providers);
+            if (gitHubUrl.isPresent() && gitRepository.isPresent() && isNotEmpty(gitCommitId)) {
+                buildScan.link("GitHub source", gitHubUrl.get() + "/" + gitRepository.get() + "/tree/" + gitCommitId);
+            } else if (isNotEmpty(gitRepo) && isNotEmpty(gitCommitId)) {
+                Optional<URI> gitRepoUri = toUri(gitRepo);
+                gitRepoUri.ifPresent(r -> {
+                    String gitRepoHost = r.getHost();
+                    String gitRepoPath = r.getPath().endsWith(".git") ? r.getPath().substring(0, r.getPath().length() - 4) : r.getPath();
+                    if (gitRepoHost.contains("github")) {
+                        buildScan.link("GitHub source", "https://" + gitRepoHost + gitRepoPath + "/tree/" + gitCommitId);
```
etiennestuder, line 409: We should construct the link via a URI object rather than by stitching the URL together by manually, since we currently assume port 80 and https protocol and no user info. Please address this in a separate PR.

etiennestuder (changes requested):
We are down to one last item:
https://github.com/gradle/common-custom-user-data-gradle-plugin/pull/150/files#r1158059776

src/main/java/com/gradle/CustomBuildScanEnhancements.java
tylerbertrand: Pushed a commit addressing this - not sure if it's exactly what you were going for though

tylerbertrand: Pushed another commit making this more concise

tylerbertrand:
@etiennestuder So unfortunately I had a pretty critical miss - when I was going through the final testing, it turns out that `URI` cannot parse the typical SSH url you'd get from a GitHub repo (for example `git@github.com:gradle/common-custom-user-data-gradle-plugin.git`).

@erichaagdev and I were discussing ways we could handle this, and we could not come up with what we felt were any great solutions. I started out with 547723c that would attempt to convert these to valid URIs with some string find/replaces.

For the typical ssh url given by GitHub when cloning ssh repositories, (e.g. `git@github.com:gradle/common-custom-user-data-gradle-plugin.git`), prepending `ssh://` and converting the `:` to a `/` makes this valid.

If your ssh url already begins with `ssh://`, it must already use a `/` to delineate and path. Running `git clone ssh://git@github.com:gradle/common-custom-user-data-gradle-plugin.git` fails, while `git clone ssh://git@github.com/gradle/common-custom-user-data-gradle-plugin.git` succeeds. This case already parses into a valid `URI`, so no action is needed.

If your ssh url begins with `ssh://`, and thus already has the correct `/` delineating host and path, it may still have a `:` in the case where the port number is specified. This case also parses into a valid `URI`, so no action is needed.

This solution does appear to work for all the cases I could think of for ssh urls returned by `git config --get remote.origin.url` (see the ssh variations mentioned above and those I tested in the PR description), but it does seem quite fragile. I wouldn't be surprised if there were cases that would break this. One obvious way to break this (though I have no idea why one would ever do this), is to manually set your origin url to some bad value, with `git remote set-url origin sshttps://somebadurl:22:asdfasdf`. 

I left this solution in the commit in the history in case you wanted to take a look. 

We discussed further and thought that it may be a better idea to go with a simpler but more resilient approach. This would cost us some functionality, but seems to be much less prone to producing a broken link. This is implemented in commit dc8201e and replaced the previous solution above. In short, it checks for `git@github.com:` or `git@gitlab.com:`, and simply replaces that with `ssh://git@github.com/` or `ssh://git@gitlab.com/`, which converts them into valid `URI`s. This is quite predictable and safe, but we lose the ability to generate links to repositories cloned using ssh from GitHub Enterprise installations (e.g. `git@github.somecompany.com` will not have a link generated for it). This second solution could be extended to include such urls, but this might point to going back to the original regex-based solutions. Definitely wanted to get your thoughts before heading that way again.

etiennestuder (approved):
LGTM. I pushed [one small refactor](https://github.com/gradle/common-custom-user-data-gradle-plugin/commit/9db721a07c9fb0d666cbfb6296f90463384cb3ef) that does not change the logic.

When you merge to main, please _squash_ the changes.

erichaagdev (approved):

PR #156, opened 2023-03-29
Add release notes to plugin description

### LINKED ISSUE #134: User can read release notes in the plugin portal
We should use the `description` field of the plugin declaration to specify release notes. Currently, it is always a static string. 

The `description` field is shown on the plugin portal for each published version. Setting the `description` field to release notes means they can be seen directly in the plugin portal, rather than only on GitHub. This makes release notes and changes much more discoverable for users wanting to update.

An example of this can be seen on the plugin portal page for `com.gradle.plugin-publish`: https://plugins.gradle.org/plugin/com.gradle.plugin-publish

tylerbertrand:
Resolves #134

Allows users to view release notes from the plugin portal and gives them an idea of what will be gained by updating to that version.

The contents of the release notes file are read in a way that supports configuration caching using `Providers.fileContents()`. This functionality was extracted into its own method, and setting `githubRelease.body` has been updated to use this. 

Configuration cache compatibility was verified using the following steps...
1. Run `./gradlew clean`
2. Run `./gradlew generatePomFileForCommonCustomUserDataPluginMarkerMavenPublication` to generate the `build/publications/commonCustomUserDataPluginMarkerMaven/pom-default.xml` file containing the release notes in `release/changes.md`. The logs indicate a new configuration cache entry has been stored.
3. Run `./gradlew generatePomFileForCommonCustomUserDataPluginMarkerMavenPublication` again without changes. The logs indicate the configuration cache entry has been re-used. The contents of `pom-default.xml` have not changed.
4. Make changes to `release/changes.md`
5. Re-run `./gradlew generatePomFileForCommonCustomUserDataPluginMarkerMavenPublication`. The logs indicate a new configuration cache entry has been stored. A new `pom-default.xml` has been generated with a description reflecting the changes to `release/changes.md`.

Also corrected `githubRelease` closure indent from 3 spaces to 4. All other indentation is 4 spaces.

[CCUD plugin in the staging portal](https://plugins.grdev.net/plugin/com.gradle.common-custom-user-data-gradle-plugin)

image

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -38,16 +38,12 @@ gradlePlugin {
 
     automatedPublishing = true
 
-    def releaseNotes = layout.projectDirectory.file('release/changes.md').asFile.text
+    def releaseNotesFile = layout.projectDirectory.file('release/changes.md')
     plugins {
         commonCustomUserData {
             id = 'com.gradle.common-custom-user-data-gradle-plugin'
             displayName = 'Gradle Enterprise Common Custom User Data Gradle Plugin'
-            description = """\
-                          |A Gradle plugin to capture common custom user data used for Gradle Build Scans in Gradle Enterprise
-                          |
-                          |$releaseNotes
-                          """.stripMargin().trim()
+            description = providers.fileContents(releaseNotesFile).asText.get().trim()
             implementationClass = 'com.gradle.CommonCustomUserDataGradlePlugin'
             tags.addAll('android', 'java', 'gradle enterprise')
         }
--- a/build.gradle
+++ b/build.gradle
@@ -39,12 +39,11 @@ gradlePlugin {
 
     automatedPublishing = true
 
-    def releaseNotesFile = layout.projectDirectory.file('release/changes.md')
     plugins {
         commonCustomUserData {
             id = 'com.gradle.common-custom-user-data-gradle-plugin'
             displayName = 'Gradle Enterprise Common Custom User Data Gradle Plugin'
-            description = providers.fileContents(releaseNotesFile).asText.get().trim()
+            description = releaseNotes().get()
             implementationClass = 'com.gradle.CommonCustomUserDataGradlePlugin'
             tags.addAll('android', 'java', 'gradle enterprise')
         }
@@ -105,4 +104,9 @@ def gitHubReleaseName() {
 
 def gitReleaseTag() {
     return "v${version}"
+}
+
+def releaseNotes() {
+    def releaseNotesFile = layout.projectDirectory.file('release/changes.md')
+    return providers.fileContents(releaseNotesFile).asText.map { it -> it.trim() }
 }
--- a/build.gradle
+++ b/build.gradle
@@ -72,16 +72,16 @@ signing {
 }
 
 githubRelease {
-   token = System.getenv('CCUD_GIT_TOKEN') ?: ''
-   owner = 'gradle'
-   repo = 'common-custom-user-data-gradle-plugin'
-   targetCommitish = 'main'
-   releaseName = gitHubReleaseName()
-   tagName = gitReleaseTag()
-   prerelease = false
-   overwrite = false
-   generateReleaseNotes = false
-   body = releaseNotes()
+    token = System.getenv('CCUD_GIT_TOKEN') ?: ''
+    owner = 'gradle'
+    repo = 'common-custom-user-data-gradle-plugin'
+    targetCommitish = 'main'
+    releaseName = gitHubReleaseName()
+    tagName = gitReleaseTag()
+    prerelease = false
+    overwrite = false
+    generateReleaseNotes = false
+    body = releaseNotes()
 }
 
 def createReleaseTag = tasks.register('createReleaseTag', CreateGitTag) {
--- a/build.gradle
+++ b/build.gradle
@@ -81,7 +81,7 @@ githubRelease {
    prerelease = false
    overwrite = false
    generateReleaseNotes = false
-   body = layout.projectDirectory.file('release/changes.md').asFile.text.trim()
+   body = releaseNotes()
 }
 
 def createReleaseTag = tasks.register('createReleaseTag', CreateGitTag) {
--- a/build.gradle
+++ b/build.gradle
@@ -38,11 +38,16 @@ gradlePlugin {
 
     automatedPublishing = true
 
+    def releaseNotes = layout.projectDirectory.file('release/changes.md').asFile.text
     plugins {
         commonCustomUserData {
             id = 'com.gradle.common-custom-user-data-gradle-plugin'
             displayName = 'Gradle Enterprise Common Custom User Data Gradle Plugin'
-            description = 'A Gradle plugin to capture common custom user data used for Gradle Build Scans in Gradle Enterprise'
+            description = """\
+                          |A Gradle plugin to capture common custom user data used for Gradle Build Scans in Gradle Enterprise
+                          |
+                          |$releaseNotes
+                          """.stripMargin().trim()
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -38,11 +38,16 @@ gradlePlugin {
 
     automatedPublishing = true
 
+    def releaseNotes = layout.projectDirectory.file("release/changes.md").asFile.text
```
etiennestuder, line 41: Is this change build caching compatible, meaning changes to the release notes will retrigger any required tasks to rerun (if any)?

I'm not saying there are cacheable tasks depending on this input, but maybe there are, and I want to be sure it was considered. (it needs to be muscle-memory to see any build change in the context of build caching and configuration caching)

etiennestuder, line 41: Single quotes suffice.

etiennestuder (changes requested):

build.gradle
tylerbertrand: Fixed

tylerbertrand: Good point - it looks like all tasks that need to be re-executed when `changes.md` changes are re-executed. So I think we are good here.

build.gradle
```diff
@@ -38,11 +38,16 @@ gradlePlugin {
...
     plugins {
         commonCustomUserData {
             id = 'com.gradle.common-custom-user-data-gradle-plugin'
             displayName = 'Gradle Enterprise Common Custom User Data Gradle Plugin'
-            description = 'A Gradle plugin to capture common custom user data used for Gradle Build Scans in Gradle Enterprise'
+            description = """\
+                          |A Gradle plugin to capture common custom user data used for Gradle Build Scans in Gradle Enterprise
+                          |
+                          |$releaseNotes
+                          """.stripMargin().trim()
```
erichaagdev, line 50: I recommend doing two things. 

- Leverage `project.description` for the first bit. Right now we don't have a project description so that would be a nice enhancement anyway. Then in this logic you can replace hard-coded text with `project.description`. 
- Encapsulate building the release notes to a function so that it can more easily be brought over to other projects.

erichaagdev: We could also scrap adding the `project.description` to the release notes entirely. The [`com.gradle.plugin-publish`](https://plugins.gradle.org/plugin/com.gradle.plugin-publish) does not include it.

build.gradle
erichaagdev: Since this is reading a file at configuration time, then this is not compatible with configuration caching.

erichaagdev: I recommend reaching out to the build tool team to see what they recommend as far as setting release notes from a file.

erichaagdev: Yes, the tasks themselves will always re-execute but changes to the `changes.md` file will not invalidate the configuration cache. Try this:

1. Run `./gradlew clean`
2. Run `./gradlew generatePomFileForCCUD`
3. Inspect the description in `build/publications/commonCustomUserDataPluginMarkerMaven/pom-default.xml`
4. Modify `release/changes.md`
5. Run `./gradlew generatePomFileForCCUD`
6. Inspect the description in `build/publications/commonCustomUserDataPluginMarkerMaven/pom-default.xml` and you will see it did not change.

tylerbertrand: Ah I see - I had been testing this exact scenario but with `publishToMavenLocal`. When invoked with that task, the description does get updated.

build.gradle
tylerbertrand: > We could also scrap adding the project.description to the release notes entirely. The [com.gradle.plugin-publish](https://plugins.gradle.org/plugin/com.gradle.plugin-publish) does not include it.

Yeah I think let's follow the example of the `plugin-publish` plugin. It already links to the repo on GitHub anyways, so you're a click away from the README.

build.gradle
erichaagdev: Very nice. 😎 

https://ge.solutions-team.gradle.com/s/wogvc5hp3e6fs/console-log?page=1#L1

build.gradle
```diff
@@ -38,11 +39,12 @@ gradlePlugin {
 
     automatedPublishing = true
 
+    def releaseNotesFile = layout.projectDirectory.file('release/changes.md')
     plugins {
         commonCustomUserData {
             id = 'com.gradle.common-custom-user-data-gradle-plugin'
             displayName = 'Gradle Enterprise Common Custom User Data Gradle Plugin'
-            description = 'A Gradle plugin to capture common custom user data used for Gradle Build Scans in Gradle Enterprise'
+            description = providers.fileContents(releaseNotesFile).asText.get()
```
erichaagdev, line 47: It is probably still a good idea to throw a `.trim()` on here. Since you're reading the value at configuration time, it is probably fine to just append `.trim()` here rather than doing a `.map { it.trim() }`.

tylerbertrand: Yep I missed that, thanks.

build.gradle
erichaagdev, line 1: We also read the contents of `release/changes.md` to set the `githubRelease.body`. It might make sense to create a function that returns the file contents provider to be used in both places.

tylerbertrand: Maybe something like...
```Groovy
Provider<String> getReleaseNotesProvider() {
    def releaseNotesFile = layout.projectDirectory.file('release/changes.md')
    return providers.fileContents(releaseNotesFile).asText
}
```
?

erichaagdev: This is how I would do it:

```groovy
def releaseNotes() {
    def releaseNotesFile = layout.projectDirectory.file('release/changes.md')
    return providers.fileContents(releaseNotesFile).asText.map { it.trim() }
}
```

tylerbertrand: Ahh yep I need to remember to keep that in mind that `map(...)` returns another `Provider`.

tylerbertrand: Ok, pushed an update adding this 👍

build.gradle
```diff
@@ -80,7 +81,7 @@ githubRelease {
    prerelease = false
    overwrite = false
    generateReleaseNotes = false
-   body = layout.projectDirectory.file('release/changes.md').asFile.text.trim()
+   body = releaseNotes().get()
```
erichaagdev, line 84: `githubRelease.body` is actually a provider so you don't need to call `.get()` here.

tylerbertrand: Fixed, thanks.

erichaagdev (approved):
LGTM! Nice work! 👌

etiennestuder (approved):
LGTM.

I'm happy with the thoroughness that went into the testing and discussion of this PR. The result is a robust improvement.

build.gradle
```diff
@@ -42,7 +43,7 @@ gradlePlugin {
         commonCustomUserData {
             id = 'com.gradle.common-custom-user-data-gradle-plugin'
             displayName = 'Gradle Enterprise Common Custom User Data Gradle Plugin'
-            description = 'A Gradle plugin to capture common custom user data used for Gradle Build Scans in Gradle Enterprise'
+            description = releaseNotes().get()
```
etiennestuder, line 46: Is description not a Provider<String> but a String?

etiennestuder, line 46: You are reading the release notes twice. I suggest you pull this into a variable called releaseNotes that you declare above the `repositories` block.

etiennestuder, line 46: Is `description` not of type Provider<String> but a String?

etiennestuder (commented):
I left a question and a request.

build.gradle
etiennestuder: I assume the answer is NO.

tylerbertrand: That's correct - sorry, I did verify that it is in fact a string, but forgot to respond to your comment.

PR #160, opened 2023-04-10
Use configuration cache compatible provider api to read version file

tylerbertrand:
User CC-compatible `Provider.fileContents()` to read version file


build.gradle
```diff
@@ -98,15 +98,16 @@ tasks.withType(Sign).configureEach {
     notCompatibleWithConfigurationCache("$name task does not support configuration caching")
 }
 
-def gitHubReleaseName() {
-    return version.toString()
-}
-
 def gitReleaseTag() {
-    return "v${version}"
+    return version.map { it -> "v$it" }
```
erichaagdev, line 102: Does Groovy have an implicit `it` like Kotlin? If so, you could simplify this to:

```suggestion
    return version.map { "v$it" }
```

tylerbertrand: Ah yup it does

build.gradle
```diff
@@ -98,15 +98,16 @@ tasks.withType(Sign).configureEach {
...
 }
 
-def gitHubReleaseName() {
-    return version.toString()
-}
-
 def gitReleaseTag() {
-    return "v${version}"
+    return version.map { "v$it" }
 }
```
erichaagdev, line 103: I feel like `gitReleaseTag` could be removed and the contents inlined. It's only called in a single place and feels redundant.

erichaagdev:
> A few of these could probably be removed, though maybe there is some value in being resilient to the possibility of the default behavior changing.

In this case I think it makes perfect sense to set them explicitly, even if the default value is the same. This plugin controls how releases of our plugin are done and we would not want that to unexpectedly change.

build.gradle
tylerbertrand: Agreed. I'll make that change as I go through the rest of the repos too - they follow this same pattern.

tylerbertrand: Actually it is called twice, if that changes your opinion.

tylerbertrand: Personally I kinda like keeping it in that case.

build.gradle
etiennestuder, line 1: Please rebase from `main` so we get the changes from #159.

build.gradle
```diff
@@ -8,7 +8,7 @@ plugins {
...
 
 group = 'com.gradle'
-version = layout.projectDirectory.file('release/version.txt').asFile.text.trim()
+version = version()
```
etiennestuder, line 11: Extract the call to version() into a variable above, like you did for the release notes provider. I would call it `pluginVersion` to avoid any naming collision/confusion.

build.gradle
```diff
@@ -76,7 +76,7 @@ githubRelease {
     owner = 'gradle'
     repo = 'common-custom-user-data-gradle-plugin'
     targetCommitish = 'main'
-    releaseName = gitHubReleaseName()
+    releaseName = version
```
etiennestuder, line 79: Set it as a provider rather than the resolved provider value, as you do for the body/release notes.

build.gradle
```diff
@@ -98,15 +98,16 @@ tasks.withType(Sign).configureEach {
     notCompatibleWithConfigurationCache("$name task does not support configuration caching")
 }
 
-def gitHubReleaseName() {
-    return version.toString()
-}
```
etiennestuder, line 103: Either inline both gitHubReleaseName and gitReleaseTag, or none of them.

If you inline both, you can avoid the duplicate calls by getting the tagName in `createReleaseTag` from the `githubRelease` extension object. (it's surprising the git plugin does not wire the values like this by default)

build.gradle
etiennestuder: See comment above.

etiennestuder (changes requested):

build.gradle
tylerbertrand: Done 👍

build.gradle
tylerbertrand: `gitHubReleaseName()` wasn't necessarily inlined - it's just not necessary any more because `releaseName` is being assigned to a `Provider<String>` instead of a `String`

build.gradle
tylerbertrand: Done - but just so I know, what is the issue with referencing `project.version` elsewhere in the script?

build.gradle
tylerbertrand: Updated this to assign `releaseName = pluginVersion` instead of `project.version`. 

Though, `project.version` was previously being assigned to the provider returned by `version()`, so why would assigning `releaseName = project.version` use the resolved provider value instead of the provider itself? Just want to make sure I'm following your reasoning.

build.gradle
tylerbertrand: Done - the `tagName` in `createReleaseTag` now comes from the `githubRelease` extension object.

build.gradle
```diff
@@ -78,18 +79,19 @@ githubRelease {
...
     targetCommitish = 'main'
-    releaseName = gitHubReleaseName()
-    tagName = gitReleaseTag()
+    releaseName = releaseVersion
+    tagName = releaseVersion.map { "v$it" }
     prerelease = false
     overwrite = false
     generateReleaseNotes = false
     body = releaseNotes
+    dryRun = true
```
etiennestuder, line 88: Why was this added?

build.gradle
etiennestuder: Resolving given it has been inlined.

build.gradle
```diff
@@ -78,18 +79,19 @@ githubRelease {
...
 }
 
 def createReleaseTag = tasks.register('createReleaseTag', CreateGitTag) {
     // Ensure tag is created only after a successful publishing
     mustRunAfter('publishPlugins')
-    tagName = gitReleaseTag()
+    tagName = githubRelease.tagName.map { it.toString() }
```
etiennestuder, line 94: Is this the mapping with toString() needed because githubRelease.tagName is a Property<CharSequence> and CreateGitTag.tagName is a Property<String> ?

tylerbertrand: It's required because `githubRelease.tagName` is a `Property<CharSequence>` and,  without the `toString()` mapping, can't be assigned to `CreateGitTag.tagName` which is a `Property<String>`

build.gradle
tylerbertrand: Was planning on removing this before merging, I've gone ahead and removed it.

build.gradle
etiennestuder: Yep, that's why I meant. I wonder why they used CharSequence???

etiennestuder: Oh now, all my `Property<String>` and `Property<CharSequence>` got lost in my comment above due to `<` (markdown).

tylerbertrand: Ahh, gotcha. As for why they used `CharSequence`, not sure. Looking at its implementing classes, none of them outside of `String` pop out to me as super useful for setting something as simple as a tag...seems like the other implementations can be easily converted to a `String` anyways. At a quick glance nothing interesting going on in their plugin code with `CharSequence` either. 🤷

PR #166, opened 2023-04-17
Toolchains are auto-provisioned via the foojay plugin

clayburn:


etiennestuder (changes requested):
I believe you should do the same as here in terms of also defining the toolchain (and plugin) in the buildSrc folder:
https://github.com/gradle/gradle-enterprise-build-validation-scripts/blob/main/buildSrc

erichaagdev (approved):

PR #193, opened 2023-06-30
Handle embedded user info in git repo url

### LINKED ISSUE #191: GitLab source link may be invalid
We've seen build scans generated with invalid links to 'GitLab source'.

Root cause seems to be invalid handling pattern matching for URLs with embedded auth details in the `toWebRepoUri` method in `Utils.java`.

In particular.. if the git origin contains embedded user info, and the user name contains 'gitlab', then the groups in `GIT_REPO_URI_PATTERN`  capture the wrong bits of the URL.

e.g. given an origin of `https://gitlab-svc-user:secret@gitlab.self-hosted.com/foo/bar.git`, the generated 'GitLab source' link points to `https://gitlab-svc-user/secret@gitlab.self-hosted.com/foo/bar`

tylerbertrand:
Fixes #191

Update `GIT_REPO_URI_PATTERN` to properly handle embedded user info in git repository URL

### Tests
* `https://gitlab-svc-user:secret@gitlab.self-hosted.com/foo/bar.git`: https://ge.solutions-team.gradle.com/s/b54q5wdxtkef4#info
* `https://github.com/gradle/common-custom-user-data-gradle-plugin.git`: https://ge.solutions-team.gradle.com/s/yv6hrmnldvo3e#info

### DIFF
--- a/src/main/java/com/gradle/Utils.java
+++ b/src/main/java/com/gradle/Utils.java
@@ -29,7 +29,7 @@ import java.util.regex.Pattern;
 
 final class Utils {
 
-    private static final Pattern GIT_REPO_URI_PATTERN = Pattern.compile("^(?:(?:https://|git://)|(?:ssh)?.*?@)(.*?(?:github|gitlab).*?)(?:/|:[0-9]*?/|:)(.*?)(?:\\.git)?$");
+    private static final Pattern GIT_REPO_URI_PATTERN = Pattern.compile("^(?:(?:https://|git://)(?:.+:.+@)?|(?:ssh)?.*?@)(.*?(?:github|gitlab).*?)(?:/|:[0-9]*?/|:)(.*?)(?:\\.git)?$");
 
     static Optional<String> sysPropertyOrEnvVariable(String sysPropertyName, String envVarName, ProviderFactory providers) {
         Optional<String> sysProperty = sysProperty(sysPropertyName, providers);
--- a/src/test/java/com/gradle/UtilsTest.java
+++ b/src/test/java/com/gradle/UtilsTest.java
@@ -1,22 +1,64 @@
 package com.gradle;
 
-import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.ArgumentsProvider;
+import org.junit.jupiter.params.provider.ArgumentsSource;
 
 import java.net.URI;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import static com.gradle.Utils.toWebRepoUri;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
 public class UtilsTest {
-    @Test
-    public void toWebRepoUri_git() {
-        assertEquals(URI.create("https://github.com/acme-inc/my-project"), toWebRepoUri("git://github.com/acme-inc/my-project.git").get());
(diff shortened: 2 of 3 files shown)


src/main/java/com/gradle/Utils.java
```diff
@@ -29,7 +29,7 @@
 
 final class Utils {
 
-    private static final Pattern GIT_REPO_URI_PATTERN = Pattern.compile("^(?:(?:https://|git://)|(?:ssh)?.*?@)(.*?(?:github|gitlab).*?)(?:/|:[0-9]*?/|:)(.*?)(?:\\.git)?$");
+    private static final Pattern GIT_REPO_URI_PATTERN = Pattern.compile("^(?:(?:https://|git://)(?:.*?:.*?@)?|(?:ssh)?.*?@)(.*?(?:github|gitlab).*?)(?:/|:[0-9]*?/|:)(.*?)(?:\\.git)?$");
```
jprinet, line 32: Didn't try but wouldn't `.+` be simpler than `.*?` ?

jprinet (commented):
Some unit tests would help to make sure the regular expression is parsing as expected

src/main/java/com/gradle/Utils.java
tylerbertrand: Yep you're right I think that is more correct

tylerbertrand:
> Some unit tests would help to make sure the regular expression is parsing as expected

@jprinet Added tests to ensure different variations of the remote repository uri are parsed as expected

src/test/java/com/gradle/UtilsTest.java
```diff
@@ -1,22 +1,64 @@
...
+            ).collect(Collectors.toSet());
+            return host.stream().flatMap(h -> remoteRepositoryUris.stream().map(r -> Arguments.arguments(h, r)));
+        }
+    }
+
+    private static class EnterpriseWebRepoUriArgumentsProvider implements ArgumentsProvider {
+
+        @Override
+        public Stream<? extends Arguments> provideArguments(ExtensionContext context) {
+            Set<String> host = Stream.of("github", "gitlab").collect(Collectors.toSet());
```
jprinet, line 56: Very nice, don't want to be annoying but we could factorize the `Stream` creation and have only the `Set` of `remoteRepositoryUris` being different in the 2 `ArgumentsProvider` implementations.

jprinet:
> > Some unit tests would help to make sure the regular expression is parsing as expected
> 
> @jprinet Added tests to ensure different variations of the remote repository uri are parsed as expected

Thanks @tylerbertrand, that's way better like this 👏, I left a comment that you may take into account.

src/test/java/com/gradle/UtilsTest.java
```diff
@@ -1,22 +1,64 @@
...
 import static com.gradle.Utils.toWebRepoUri;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
 public class UtilsTest {
-    @Test
-    public void toWebRepoUri_git() {
-        assertEquals(URI.create("https://github.com/acme-inc/my-project"), toWebRepoUri("git://github.com/acme-inc/my-project.git").get());
-        assertEquals(URI.create("https://gitlab.com/acme-inc/my-project"), toWebRepoUri("git://gitlab.com/acme-inc/my-project.git").get());
+
+    @ParameterizedTest
```
runningcode, line 20: Good to create tests for these cases!
Can you tell me more about the decision to use a parameterized test here vs just having a list of values in the test itself?

runningcode (approved):

src/test/java/com/gradle/UtilsTest.java
tylerbertrand: No worries! So I gave it a shot, and I think I prefer it as-is, since what's being duplicated is very small, and only in two places. If further related argument providers were added, I'd probably be in favor of de-duplicating that.

PR #230, opened 2023-10-09
Do not add `maxParallelForks` default values

SimonMarquis:


jprinet:
Hi @SimonMarquis and thanks for the contribution.

We won't merge this PR as we will soon remove the `maxParallelForks` custom value from the plugin. We already received feedback that this was not useful most of the time.

We will add a [sample](https://github.com/gradle/gradle-enterprise-build-config-samples/tree/main/build-data-capturing-gradle-samples) to still have the option to capture the value.

PR #243, opened 2023-11-27
Check Access Key Presence to avoid Build Cache errors on PR builds where Key is not present

runningcode:


settings.gradle
```diff
@@ -26,7 +26,9 @@ buildCache {
 
     remote(gradleEnterprise.buildCache) {
         enabled = true
-        push = isCI
+        // Check access key presence to avoid build cache errors on PR builds when access key is not present
+        def accessKey = System.getenv("GRADLE_ENTERPRISE_ACCESS_KEY")
+        push = isCI && accessKey != null && !accessKey.isEmpty()
```
etiennestuder, line 31: This can be simplified to isCi && accessKey.

etiennestuder, line 31: If we do this for this repo, we should do it for the other repos, too. Of, if we only need/want one example, we should do it in the config samples repo (for groovy and Kotlin).

runningcode: Agreed, we should do this for all the repos.

runningcode: updated

etiennestuder: But please be sure to verify it and not take my word for it.

runningcode: Yup, this PR is from a fork, and there are no 403 errors or other remote build cache errors in the logs: https://github.com/gradle/common-custom-user-data-gradle-plugin/actions/runs/7003812403/job/19050268047

PR #250, opened 2023-12-06
Update studio version property to new version

ZacSweers:
This appears to have changed in recent studio versions to a new key


cdsap:
Hi @ZacSweers, thanks for the contribution

PR #273, opened 2024-02-26
Replace references to GE plugin classes w/ reflective proxies

pshevche:
## Summary

This is part 1 of changes required to support the Develocity plugin. GE extension + model classes won't be available starting with the next major release of the Develocity plugin. Therefore, we won't be able to reference those classes if we want to continue compiling against the latest version of the Develocity plugin. This PR replaces all references to GE classes with reflective proxies. In a follow-up PR, I will add an additional adapter layer on top of these proxies that would allow the uniform configuration of the Develocity and GE plugins.


src/main/java/com/gradle/ccud/adapters/reflection/Java8MethodHandleLookup.java
```diff
@@ -0,0 +1,33 @@
+package com.gradle.ccud.adapters.reflection;
+
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+
+class Java8MethodHandleLookup implements MethodHandleLookup {
```
pshevche, line 9: 💬 Shameless copy of the utility classes available in the Develocity plugin. Ideas how to share these classes are welcome.

jean-andre-gauthier (approved):

marcphilipp (approved):

README.md
```diff
@@ -22,7 +22,7 @@ You can leverage this plugin for your project in one of two ways:
 ## Applying the published plugin
 
 The Common Custom User Data Gradle plugin is available in the [Gradle Plugin Portal](https://plugins.gradle.org/plugin/com.gradle.common-custom-user-data-gradle-plugin). This
-plugin requires the [Develocity Gradle plugin](https://plugins.gradle.org/plugin/com.gradle.enterprise) to also be applied in your build in order to have an effect.
+plugin requires the [Develocity Gradle plugin](https://plugins.gradle.org/plugin/com.gradle.develocity) to also be applied in your build in order to have an effect.
```
runningcode, line 25: Let's not make this change quite yet since this isn't published

build.gradle
```diff
@@ -60,8 +55,36 @@ tasks.withType(ValidatePlugins).configureEach {
...
+            useJUnitJupiter()
+
+            dependencies {
+                implementation(project())
+                implementation(platform('org.junit:junit-bom:5.10.2'))
+                implementation('org.junit.jupiter:junit-jupiter')
+                implementation('org.mockito:mockito-core:4.11.0')
+                implementation('org.mockito:mockito-junit-jupiter:4.11.0')
+                implementation('com.gradle:gradle-enterprise-gradle-plugin:3.16.2') {
+                    because("Verify that reflective proxies and adapters work with actual plugin classes")
```
runningcode, line 79: Nice use of because.

src/enterpriseCompatibilityTest/java/com/gradle/ccud/adapters/enterprise/proxies/BuildScanExtensionProxyTest.java
```diff
@@ -0,0 +1,252 @@
...
+
+        // when
+        when(extension.getTermsOfServiceUrl()).thenReturn(url);
+
+        // then
+        assertEquals(url, proxy.getTermsOfServiceUrl());
+    }
+
+    @Test
+    @DisplayName("can set and retrieve the terms of service agreement using the extension proxy")
```
runningcode, line 81: I was wondering why we needed this but then I realized that while this isn't directly used, we customers who extend the functionality of CCUD to customize any of the functionality of the Develocity plugin. For example in `CustomDevelocityConfig` 👍

src/main/java/com/gradle/Overrides.java
```diff
@@ -1,23 +1,27 @@
...
- * Provide standardized Gradle Enterprise configuration. By applying the plugin, these settings will automatically be applied.
+ * Provide standardized Develocity configuration. By applying the plugin, these settings will automatically be applied.
  */
 final class Overrides {
 
-    // system properties to override Gradle Enterprise configuration
-    static final String GRADLE_ENTERPRISE_URL = "gradle.enterprise.url";
-    static final String GRADLE_ENTERPRISE_ALLOW_UNTRUSTED_SERVER = "gradle.enterprise.allowUntrustedServer";
+    // system properties to override Develocity configuration
+    static final String DEVELOCITY_URL = "develocity.url";
```
runningcode, line 23: I think we need to check the old `gradle.enterprise.url` as well. This is for backward compatibility with the BVS.
I think the same is true for `allowUntrustedServer` and other system properties.

README.md
```diff
@@ -70,10 +70,10 @@ to modify the build scripts. For example, to disable the local build cache when
 
 ### Develocity settings
 
-| Develocity API                        | System property                        | Environment variable                   |
-|:--------------------------------------|:---------------------------------------|:---------------------------------------|
-| gradleEnterprise.server               | gradle.enterprise.url                  | GRADLE_ENTERPRISE_URL                  |
-| gradleEnterprise.allowUntrustedServer | gradle.enterprise.allowUntrustedServer | GRADLE_ENTERPRISE_ALLOWUNTRUSTEDSERVER |
+| Develocity API                  | System property                 | Environment variable            |
+|:--------------------------------|:--------------------------------|:--------------------------------|
+| develocity.server               | develocity.url                  | DEVELOCITY_URL                  |
```
runningcode, line 75: I mentioned below but we'll need to keep the ability to set these with the old properties for compatibility with the BVS and CI injection.

runningcode (commented):
This looks great!
I added some comments. It would also be helpful that @erichaagdev examine this from the perspective of BVS compatibility and @clayburn examine this from the perspective of CI Injection compatibility.

runningcode (commented):
One more thought, can you tell me about why you chose the approach to use proxy classes as opposed to using the Develocity Plugin as a `compileOnly` dependency and only calling those classes when they are available at runtime?

README.md
clayburn: CI auto-injection does not use these properties to instrument Gradle builds. Instead, an init script directly configures the Develocity Gradle Plugin.

CI auto-injection does require backwards compatibility of many such properties for auto-injection of Maven builds. However, for Maven, these are handled by the Develocity Maven Extension, so they will not be in scope of any changes to Common Custom User Data Maven Extension.

I still agree with keeping backwards compatibility with the old properties for users.

clayburn (commented):
Auto-injection interacts with CCUD via:

* The Plugin ID: `com.gradle.common-custom-user-data-gradle-plugin`
* The Plugin class name: 'com.gradle.CommonCustomUserDataGradlePlugin`

From a code review perspective, given that neither of these change, these changes look good from an auto-injection point of view. I would like to see that tested to verify I am correct.

I did not give an in-depth review of the files in the `enterpriseCompatibilityTest` or `ccud/adapters`

README.md
erichaagdev: The build validation scripts are the same as the CI plugins. They configure the Develocity Gradle plugin via an init script.

src/main/java/com/gradle/Overrides.java
erichaagdev: The build validation scripts do not rely on these overrides. They configure the Develocity Gradle plugin directly via an init script.

erichaagdev (approved):

build.gradle
pshevche: Thank you :)

src/enterpriseCompatibilityTest/java/com/gradle/ccud/adapters/enterprise/proxies/BuildScanExtensionProxyTest.java
pshevche: Exactly, that was the idea behind the entire adapter layer.

src/main/java/com/gradle/CustomDevelocityConfig.java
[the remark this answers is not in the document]
pshevche: Thank you for spotting this! Fixed in https://github.com/gradle/common-custom-user-data-gradle-plugin/pull/273/commits/54f230384e68f5e2b3110a365f8ee36f6426c3bf

README.md
pshevche: That was the idea, but now I think about merging PRs before that. With all the CI plugin updates, I just have too many moving parts on my plate, and would like to close them one-by-one.

So, I reverted this change for now in https://github.com/gradle/common-custom-user-data-gradle-plugin/pull/273/commits/ca43d566bd4ca62dbf66a167b1b3e0b41df657dd

src/main/java/com/gradle/Overrides.java
pshevche: I restored old properties, and we will fall back to those if DV properties are not set. See https://github.com/gradle/common-custom-user-data-gradle-plugin/pull/273/commits/fc08a7d1f36eae6bdc1c24378c3c25321a00d37a

README.md
pshevche: I now fall back to the GE props in https://github.com/gradle/common-custom-user-data-gradle-plugin/pull/273/commits/fc08a7d1f36eae6bdc1c24378c3c25321a00d37a. But I think it is better to not document those and encourage people to use new properties instead.

runningcode (approved):

PR #274, opened 2024-02-26
Add an adapter layer for interacting with the Develocity and Gradle Enterprise plugins

pshevche:
## Summary

This is a follow-up PR for https://github.com/gradle/common-custom-user-data-gradle-plugin/pull/273. It adds an adapter layer on top of the proxy layer, which allows interacting with both the Develocity and Gradle Enterprise plugins in a uniform way. In the final PR of this series, I'll update the plugin classes to use the adapter layer and configure any extension that it finds.

### DIFF
--- a/src/main/java/com/gradle/ccud/adapters/BuildScanObfuscationAdapter.java
+++ b/src/main/java/com/gradle/ccud/adapters/BuildScanObfuscationAdapter.java
@@ -1,0 +1,17 @@
+package com.gradle.ccud.adapters;
+
+import java.net.InetAddress;
+import java.util.List;
+import java.util.function.Function;
+
+/**
+ * Adapter for {@link com.gradle.develocity.agent.gradle.scan.BuildScanDataObfuscationConfiguration} and {@link com.gradle.ccud.adapters.enterprise.proxies.BuildScanDataObfuscationProxy}
+ */
+public interface BuildScanObfuscationAdapter {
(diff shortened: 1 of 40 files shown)


src/main/java/com/gradle/ccud/adapters/BuildScanObfuscationAdapter.java
```diff
@@ -0,0 +1,17 @@
+package com.gradle.ccud.adapters;
+
+import java.net.InetAddress;
+import java.util.List;
+import java.util.function.Function;
+
+/**
+ * Adapter for {@link com.gradle.develocity.agent.gradle.scan.BuildScanDataObfuscationConfiguration} and {@link com.gradle.ccud.adapters.enterprise.proxies.BuildScanDataObfuscationProxy}
+ */
+public interface BuildScanObfuscationAdapter {
```
pshevche, line 10: 💬 Some adapter interfaces are the same as proxy interfaces. We can use just one, but I found the duplication helpful in this case. I got lost a couple of times in the proxy logic when the interface is used both as the adapter and the proxy. This also can be more easily enforced with architecture tests, which will be added in a later PR. Let me know if you feel strongly about de-duplicating interfaces, and I can revisit this.

marcphilipp (approved):

runningcode (approved):

PR #277, opened 2024-02-27
Add capturing the CI provider

ribafish:
Add capturing the CI provider as a custom value named `CI provider`

### DIFF
--- a/src/main/java/com/gradle/CustomBuildScanEnhancements.java
+++ b/src/main/java/com/gradle/CustomBuildScanEnhancements.java
@@ -182,6 +182,7 @@ final class CustomBuildScanEnhancements {
         @Override
         public void execute(BuildResult buildResult) {
             if (isJenkins(providers) || isHudson(providers)) {
+                String ciProvider = isJenkins(providers) ? "Jenkins" : "Hudson";
                 String controllerUrlEnvVar = isJenkins(providers) ? "JENKINS_URL" : "HUDSON_URL";
 
                 Optional<String> buildUrl = envVariable("BUILD_URL", providers);
@@ -191,6 +192,7 @@ final class CustomBuildScanEnhancements {
                 Optional<String> stageName = envVariable("STAGE_NAME", providers);
                 Optional<String> controllerUrl = envVariable(controllerUrlEnvVar, providers);
 
+                buildScan.value("CI provider", ciProvider);
                 buildUrl.ifPresent(url ->
                         buildScan.link(isJenkins(providers) ? "Jenkins build" : "Hudson build", url));
                 buildNumber.ifPresent(value ->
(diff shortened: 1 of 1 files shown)


src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -182,6 +182,7 @@ private CaptureCiMetadataAction(BuildScanExtension buildScan, ProviderFactory pr
         @Override
         public void execute(BuildResult buildResult) {
             if (isJenkins(providers) || isHudson(providers)) {
+                String ciProvider = isJenkins(providers) ? "Jenkins" : "Hudson";
```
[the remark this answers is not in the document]
ribafish: I thought about doing this in a separate function, for example, but at the end, I opted for the way I did it as all the other values determined by the CI provider are set per CI provider in each of their own ifs.

src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -191,6 +192,7 @@ public void execute(BuildResult buildResult) {
                 Optional<String> stageName = envVariable("STAGE_NAME", providers);
                 Optional<String> controllerUrl = envVariable(controllerUrlEnvVar, providers);
 
+                buildScan.value("CI provider", ciProvider);
```
runningcode, line 195: We typically don't have constants for custom value keys but it might make sense here because this one is repeated.

src/main/java/com/gradle/CustomBuildScanEnhancements.java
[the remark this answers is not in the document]
runningcode: I see you point as well. Let's let others weigh in on this.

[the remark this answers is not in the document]
jprinet: Either would be ok to me 🤷‍♂️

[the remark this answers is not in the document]
etiennestuder: For my taste, I would keep it as Gasper currently has it. It keeps all value assignments in one place.CI.

src/main/java/com/gradle/CustomBuildScanEnhancements.java
etiennestuder: For my taste, I would not add a constant here.

src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -284,6 +290,7 @@ public void execute(BuildResult buildResult) {
             }
 
             if (isGitLab(providers)) {
+                buildScan.value("CI provider", "GitLab CI");
```
etiennestuder, line 293: I would take away the `CI` part.

etiennestuder (changes requested):

src/main/java/com/gradle/CustomBuildScanEnhancements.java
ribafish: Just to be clear, make it only `GitLab`, but keep the `CI provider` value key, correct?

etiennestuder: @ribafish Yes.

runningcode (approved):

erichaagdev (approved):
LGTM. Really straightforward. 👍

PR #280, opened 2024-03-04
Replace proxy interfaces with plugin classes for GE API

pshevche:
## Summary
To align the CCUD plugin with the CCUD extension, we now also compile against both the GE and the DV plugins. To avoid dependency conflicts, adapter layers for GE and DV are now in isolated source sets. Each source set depends only on the plugin artifact it needs.

The new plugin structure:
* `compatibilityApi`: contains the adapter interfaces which are shared by the GE and DV adapter source sets, and are used by the plugin and end users;
* `(enterprise|develocity)Compatibility`: implement adapter interfaces for GE/DV APIs respectively;
* `(enterprise|develocity)CompatibilityTest`: tests for ☝️;
* `main`: entry point for the plugin; `AdapterFactory` class decides which adapter layer to use.


src/compatibilityApi/java/com/gradle/ccud/adapters/DevelocityAdapter.java
```diff
@@ -0,0 +1,39 @@
...
+
+import org.gradle.api.Action;
+import org.gradle.caching.configuration.AbstractBuildCache;
+
+import javax.annotation.Nullable;
+
+/**
+ * Adapter for {@link com.gradle.develocity.agent.gradle.DevelocityConfiguration} and {@link com.gradle.ccud.adapters.enterprise.proxies.GradleEnterpriseExtensionProxy}
+ */
+public interface DevelocityAdapter {
```
runningcode, line 11: Is this DevelocityApiAdapter in the Maven extension?

build.gradle
```diff
@@ -26,8 +26,27 @@ repositories {
...
+
+    develocityCompatibilityCompileOnly(gradleApi())
+    develocityCompatibilityCompileOnly('com.gradle:develocity-gradle-plugin:3.17-rc-1')
+    develocityCompatibilityImplementation(sourceSets.compatibilityApi.output)
+
+    enterpriseCompatibilityCompileOnly(gradleApi())
+    enterpriseCompatibilityCompileOnly('com.gradle:gradle-enterprise-gradle-plugin:3.16.2')
+    enterpriseCompatibilityImplementation(sourceSets.compatibilityApi.output)
+
+    api(sourceSets.compatibilityApi.output)
```
runningcode, line 46: I'm curious about why we are exposing the `enterpriseCompatibility` classes and `develocityCompatibility` classes as `api`. I understand why the `compatibilityApi` is marked as `api` but why does the consumer also need to know about the implementation details here?

src/compatibilityApi/java/com/gradle/ccud/adapters/ProxyFactory.java
```diff
@@ -94,10 +74,6 @@ private static Spec<Object> adaptSpecArg(Spec spec) {
             return arg -> spec.isSatisfiedBy(createLocalProxy(arg));
         }
 
-        private static Object readAnnotationValueInConfigurationCacheCompatibleWay(Annotation annotation) throws IllegalAccessException, InvocationTargetException, NoSuchMethodException {
-            return annotation.getClass().getMethod("value").invoke(annotation);
-        }
-
         @SuppressWarnings({"rawtypes", "unchecked"})
         private static Function<Object, Object> adaptFunctionArg(Function func) {
             return arg -> func.apply(createLocalProxy(arg));
```
runningcode, line 79: are we still using the other methods in this class like `createLocalProxy`?

src/enterpriseCompatibilityTest/java/com/gradle/ccud/adapters/enterprise/BuildScanDataObfuscationAdapterTest.java
```diff
@@ -34,7 +33,7 @@ class BuildScanDataObfuscationAdapterTest {
     @BeforeEach
     void setup() {
         obfuscation = mock();
-        adapter = new BuildScanDataObfuscationAdapter(ProxyFactory.createProxy(obfuscation, BuildScanDataObfuscationProxy.class));
+        adapter = new BuildScanDataObfuscationAdapter(ProxyFactory.createProxy(obfuscation, BuildScanDataObfuscation.class));
```
runningcode, line 36: just for my understanding, why do we need the proxy here? can the adapter directly use the class directly?

runningcode (commented):
This looks good and I like this approach since it reduces the amount of proxy classes and makes it more similar to the approach used in the CCUD Maven extension. I had two questions.
1. I thought this would be able to get rid of all the proxies. There are still some proxies needed, can you help me understand why?
2. Not related to this PR, but is there a way we can be more strict about the version of CCUD used in the tests? There's the develocity test and the gradle enterprise test. I'm worried about the Develocity version used in the build itself leaking in to the classpath of the test and upgrading/downgrading it and leading to incorrect test results. Is that possible? right now the GE Plugin used for the build is lower than the `DV` compatibility test, if it is higher than the `GE` compatibility test, will it upgrade that test?

src/compatibilityApi/java/com/gradle/ccud/adapters/DevelocityAdapter.java
pshevche: This is the `CoreApiAdapter`. Finding a suitable name for the root adapter is a bit tricky :)

src/enterpriseCompatibilityTest/java/com/gradle/ccud/adapters/enterprise/BuildScanDataObfuscationAdapterTest.java
pshevche: We don't necessarily use it in the test, but an adapter will be created in reality. So, I wanted tests to also go through proxy methods.

pshevche:
> 1. I thought this would be able to get rid of all the proxies. There are still some proxies needed, can you help me understand why?

We still need proxies because CCUD and DV/GE plugins are not necessarily loaded by the same classloaders, and we need to access them via reflection. This has been like this for a couple of releases now. I think this happens especially often if one of the plugins comes from an init script.

> 2. Not related to this PR, but is there a way we can be more strict about the version of CCUD used in the tests? There's the develocity test and the gradle enterprise test. I'm worried about the Develocity version used in the build itself leaking in to the classpath of the test and upgrading/downgrading it and leading to incorrect test results. Is that possible? right now the GE Plugin used for the build is lower than the `DV` compatibility test, if it is higher than the `GE` compatibility test, will it upgrade that test?

I don't think that the plugin applied to the build will leak to the test classpath in any way. Now, each compatibility test source set declares an explicit dependency on the plugin under test. And that version will be used.

build.gradle
pshevche: Fair, changed it in https://github.com/gradle/common-custom-user-data-gradle-plugin/pull/280/commits/1e1951475bfa44958ea75c337d6d078876d6568c

jean-andre-gauthier (approved):

src/compatibilityApi/java/com/gradle/ccud/adapters/ProxyFactory.java
runningcode: Ah I see, 👍

runningcode:
> Now, each compatibility test source set declares an explicit dependency on the plugin under test. And that version will be used.
yes but versions can be overridden or upgraded if they are not locked to a specific version. I'm not an expert but if you are confident that this won't happen then I'll approve.

runningcode (approved):

PR #286, opened 2024-03-19
Replace custom adapters with public adapters library

pshevche:
## Summary

See https://github.com/gradle/develocity-agent-adapters


jean-andre-gauthier (approved):

jprinet (approved):
LGTM

build.gradle
```diff
@@ -64,10 +45,13 @@ java {
     }
 }
 
-jar {
-    from sourceSets.compatibilityApi.output
-    from sourceSets.develocityCompatibility.output
-    from sourceSets.enterpriseCompatibility.output
+shadowJar {
+    // required by the plugin-publish-plugin
+    archiveClassifier = ''
```
runningcode, line 50: do the regular `jar` task and the `shadow` jar task now produce output to the same location? @pshevche

pshevche: Yes, they do

runningcode: Can we fix that? Overlapping outputs are a problem for correctness/reproducability. We will either have a `regular jar` or a `shadow jar` depending on the ordering of the tasks in the build.

pshevche: Do you know how to fix this? I can look into this with some pointers.

runningcode: I see two possibilities, 1. give it a different classifier, 2. disable the jar task. Maybe the BT team has more ideas though.

PR #287, opened 2024-03-21
Disable the default JAR task in favor of the shadow JAR task

pshevche:
## Summary

Following up on https://github.com/gradle/common-custom-user-data-gradle-plugin/pull/286#discussion_r1531851374.
I did not get any suggestions from the BT team, so I went with the other alternative @runningcode suggested.


jprinet (approved):
My understanding is that the `jar.enabled = false` prevents the Jar generation, thus the overlapping output situation 👍

PR #298, opened 2024-04-05
Migrate build logic to Kotlin DSL

3flex:
Closes #186 

The diff from GitHub makes it hard to review, but the commit history should be clear.

PR #299, opened 2024-04-09
Defer token provider and migrate to com.gradleup.shadow

Goooler:
Follow up #298.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -5,7 +5,7 @@ plugins {
     id("com.gradle.plugin-publish") version "1.2.1"
     id("com.github.breadmoirai.github-release") version "2.5.2"
     id("org.gradle.wrapper-upgrade") version "0.11.4"
-    id("com.github.johnrengelman.shadow") version "8.1.1"
+    id("com.gradleup.shadow") version "8.3.0"
 }
 
 val releaseVersion = releaseVersion()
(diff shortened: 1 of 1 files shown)


erichaagdev:
Hi, @Goooler. Thank you for your contribution.

Can you revert the changes to `settings.gradle.kts`? For now, we've agreed as a team to keep with the eager APIs there. We will re-evaluate that decision once Gradle's remote build cache API migrates to using providers. Our preference is to use providers wherever possible, but it doesn't make sense to use `providers.environmentVariable(...)` when calling `.isPresent` immediately resolves it.

The changes in `build.gradle.kts` look fine. 👌

build.gradle.kts
```diff
@@ -5,7 +5,7 @@ plugins {
     id("com.gradle.plugin-publish") version "1.2.1"
     id("com.github.breadmoirai.github-release") version "2.5.2"
     id("org.gradle.wrapper-upgrade") version "0.11.4"
-    id("com.github.johnrengelman.shadow") version "8.1.1"
+    id("com.gradleup.shadow") version "8.3.0"
```
Goooler, line 8: Also migrated this, see https://github.com/GradleUp/shadow/releases/tag/8.3.0.

erichaagdev:
Merged. Thanks again for your contribution @Goooler.

Issue #311, opened 2024-05-31
Add new tag to identify scans of tasks debugging from the IDE


rolgalan:
Hi! 👋 

When a task is executed from the IDE with the debugger enabled, the total build time reported will include all the debugging.

While this is expected, it might introduce some artificial outliers in the data. When looking into the p95, searching for long builds or aggregating results this might induce to a wrong analysis of the build health of a project.

I detected this by debugging some unit tests, which I believe is a pretty common use case when developing.

I noticed that IntellijIdea (checked with Android Studio) attaches 2 environment properties when debugging that are not present in normal runs:

```
idea.debugger.dispatch.port : 57328
idea.debugger.dispatch.addr : 127.0.0.1
```

Any of these properties can be used to add a new tag like `IDE debugging` to identify (and filter out when required) these scans.

I've added a draft PR #312 with a proposal to solve this. Let me know if you want to move it forward with this improvement :)

Thanks!

erichaagdev:
Hi, @rolgalan. Thanks for taking the time to create this issue and corresponding PR.

Having an indicator that a build was using a debugger would be a good addition to CCUD. 

We need to ensure the solution will work in all cases. It needs to work for the sibling [CCUD Maven extension](https://github.com/gradle/common-custom-user-data-maven-extension) and other IDEs. It's also possible to attach a debugger to the build itself.

After taking a quick look at your PR, I don't see any reason your changes couldn't be ported to the Maven extension, but they don't support debugging from Eclipse. 

We will continue having discussions internally about the direction we want to take with these changes and report back.

rolgalan:
I don't work with any Maven project, so I was not even aware of this sibling extension.

Thanks for looking into it. Let me know if you want that I open the PR to review or keep it in draft/discard it.

PR #312, opened 2024-05-31
Add new `IDE Debugging` tag (solves #311)

rolgalan:
This solves #311 to help identifying builds executed with debugging enabled.

Tested in Android Studio Jellyfish | 2023.3.1 Patch 1.


erichaagdev:
Hi, @rolgalan. We are still interested in implementing this feature, but unfortunately, we haven't been able to prioritize the necessary work. As I mentioned in #311, we need to ensure this works for all cases and tackle this holistically across the entire CCUD portfolio (we even have [sbt](https://github.com/gradle/common-custom-user-data-sbt-plugin) now). This is something we will need to do ourselves. 

We appreciate the work you put into this, but for now, I will close the PR as it has been open for nearly 2 years. We are still tracking the issue you created to monitor the implementation of this.

Thanks again for your contributions. ❤️

PR #315, opened 2024-06-14
Always use the new Develocity API even if plugin applied using old ID

pshevche:
## Summary

Both `com.gradle.develocity` and `com.gradle.enterprise` callbacks will be invoked by Gradle since the Develocity plugin transitively applies the GE plugin. The order in which these callbacks are called can vary, leading to the CCUD plugin using the deprecated API even if the user applied the plugin using the new ID. This PR suggests to check if the new API is present regardless of which plugin ID was used and which callback was triggered first. If it is available, the CCUD will always use the new API.


runningcode (approved):

pshevche:
@runningcode , I lost my write access to the CCUD repo. Please merge whenever you feel comfortable.

runningcode:
Thanks for the fix @pshevche

PR #317, opened 2024-06-20
Github actions access token migration

ribafish:
- Migrated GitHub actions gradle executions to access token auth
- Migrated GitHub actions cross build tests of BVS env var access key to develocity naming


.github/workflows/submit-github-dependency-graph.yml
```diff
@@ -19,4 +19,4 @@ jobs:
     - name: Setup Gradle
```
erichaagdev, line 19:
```suggestion
    - name: Submit dependency graph
```

erichaagdev (changes requested):

PR #331, opened 2024-08-30
NOTICE file is added and included in all assembled JAR files along with LICENSE

erichaagdev:
This PR adds a NOTICE file to the root of the project. The NOTICE file includes the licenses of all runtime dependencies, as can be seen in [this build scan](https://ge.solutions-team.gradle.com/s/rh6xon4ychuli/dependencies?toggled=W1swXSxbMCwyXV0).

Both the NOTICE and LICENSE files are then included in all assembled JAR files.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -47,6 +47,13 @@ tasks.shadowJar {
     archiveClassifier = ""
 }
 
+tasks.withType<Jar>().configureEach {
+    into(".") {
+        from(layout.projectDirectory.file("LICENSE"))
+        from(layout.projectDirectory.dir("release/distribution"))
+    }
+}
+
 gradlePlugin {
     website = "https://github.com/gradle/common-custom-user-data-gradle-plugin"
     vcsUrl = "https://github.com/gradle/common-custom-user-data-gradle-plugin.git"
--- a/release/distribution/NOTICE
+++ b/release/distribution/NOTICE
@@ -1,0 +1,36 @@
+The following copyright statements and licenses apply to various third party open
+source software packages (or portions thereof) that are distributed with
+this content.
+
+TABLE OF CONTENTS
+=================
+
+The following is a listing of the open source components detailed in this
+document.  This list is provided for your convenience; please read further if
+you wish to review the copyright notice(s) and the full text of the license
+associated with each component.
+
+
+**SECTION 1: Apache License, V2.0**
+  * Develocity Gradle Plugin Adapters
+    * com.gradle:develocity-gradle-plugin-adapters
+
+SECTION 1: Apache License, V2.0
+================================
+
+Develocity Gradle Plugin Adapters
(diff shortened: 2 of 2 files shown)


erichaagdev:
### Sources

image

### Javadoc

image

### Plugin

image

release/distribution/NOTICE
```diff
@@ -0,0 +1,36 @@
...
+
+The following is a listing of the open source components detailed in this
+document.  This list is provided for your convenience; please read further if
+you wish to review the copyright notice(s) and the full text of the license
+associated with each component.
+
+
+**SECTION 1: Apache License, V2.0**
+  * Develocity Maven Extension Adapters
+    * com.gradle:develocity-maven-extension-adapters
```
etiennestuder, line 16: Something Maven related in the Gradle plugin?

etiennestuder (changes requested):

release/distribution/NOTICE
erichaagdev: Great catch. It should be `com.gradle:develocity-gradle-plugin-adapters`. I must have only noticed `adapters` and overlooked the build tool. 

https://ge.solutions-team.gradle.com/s/rh6xon4ychuli/dependencies?focusedDependency=WzAsMiwxLFswLDIsWzFdXV0&toggled=W1swXSxbMCwyXV0

etiennestuder: Got it. I assume you will adjust all occurrences. No need for me to re-review.

erichaagdev: Fixed – but I want to wait to merge until I can verify all JARs include the expected files when published to Maven local, however [I can't right now](https://gradle.slack.com/archives/C04U1H825D5/p1725375229470849).

build.gradle.kts
```diff
@@ -47,6 +47,13 @@ tasks.shadowJar {
     archiveClassifier = ""
 }
 
+tasks.withType<Jar> {
```
etiennestuder, line 50: Do you need to use `configure.each` to avoid early task instantiation?
https://docs.gradle.org/current/userguide/lazy_configuration.html#from_a_build_script

etiennestuder (approved):

etiennestuder (changes requested):
Waiting for Eric to reply to my comment before approving.

build.gradle.kts
erichaagdev: You are correct. Fixed in most recent changes. 82204a7e67fe67face714da6d453e63ed8f68e3d

release/distribution/NOTICE
erichaagdev: Fixed – but I want to wait to merge until I can verify all JARs include the expected files when published to Maven local, however I can't right now because the publish task fails. This is being discussed via Slack internally.

PR #352, opened 2024-11-18
Add workflow for cross-version testing of current and past plugin versions

bigdaz:
This PR adds a new "Check cross-version compatibility" workflow that can be used to verify the compatibility of a selected CCUD Gradle Plugin with a range of Develocity, Gradle Enterprise and Build Scan plugins. 

At this stage, the only functionality that is tested is that, when the CCUD plugin is applied:
- The build succeeds
- A build scan can be published

No attempt is (yet) made to verify that CCUD features are working correctly.

### DIFF
--- a/.github/workflows/compatibility-check-cross-version.yml
+++ b/.github/workflows/compatibility-check-cross-version.yml
@@ -4,6 +4,7 @@ on:
   workflow_dispatch:
     inputs:
       ccud-version:
+        description: The CCUD version to test. Leave blank to build and test CCUD from sources.
         required: false
         type: string
 
@@ -75,7 +76,7 @@ jobs:
       fail-fast: false
       matrix:
         ccud-plugin-version: ${{ fromJSON(format('[ ''{0}'' ]', needs.prepare.outputs.test-ccud-version )) }}
-        ge-plugin-version: ["3.18.2", "3.17.6", "3.16.2", "3.15.1", "3.14.1", "3.13.4", "3.12.6", "3.11.4", "3.10.3", "3.9", "3.8.1", "3.7.2", "3.6.4", "3.5.2", "3.4.1", "3.3.4", "3.2.1", "3.1.1", "3.0"]
+        ge-plugin-version: ["3.18.2", "3.17.6", "3.16.2", "3.15.1", "3.14.1", "3.13.4", "3.12.6", "3.11.4", "3.10.3", "3.9", "3.8.1", "3.7.2", "3.6.4", "3.5.2", "3.4.1", "3.3.4", "3.2.1"] ## Versions < 3.2.1 are unsupported ["3.1.1", "3.0"]
         gradle-version: [8, 5]
         include:
           - expect-failure: false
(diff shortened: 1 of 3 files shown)


.github/workflows/compatibility-check-single-version.yml
```diff
@@ -0,0 +1,84 @@
...
+        uses: actions/download-artifact@v4
+        with:
+          name: ccud-plugin-repository
+          path: build/local-plugin-repository
+
+      - name: Make local plugin repository available to samples
+        if: ${{ inputs.build-ccud-plugin }}
+        working-directory: dv-solutions/sample-projects/gradle
+        run: |
+          find . -type f -name "settings.gradle" -exec sed -i '1i pluginManagement {\n  repositories {\n    maven { url "${{ github.workspace }}/build/local-plugin-repository" }\n    gradlePluginPortal()\n  }\n}' {} +
```
erichaagdev, line 50: What is the `1i ` for prior to `pluginManagement`?

.github/workflows/compatibility-check-single-version.yml
```diff
@@ -0,0 +1,84 @@
...
+          distribution: 'temurin'
+          java-version: 8
+
+      - name: Set up Gradle
+        uses: gradle/actions/setup-gradle@main
+        with:
+          develocity-access-key: ${{ secrets.DEVELOCITY_ACCESS_KEY }}
+
+      - name: Run Gradle
+        continue-on-error: ${{ inputs.expect-failure }}
```
erichaagdev, line 68: I don't see where we verify there actually was a failure in the case of `expect-failure: true`.

.github/workflows/compatibility-check-cross-version.yml
```diff
@@ -0,0 +1,133 @@
...
+        run: 
+          echo "ccud-version=${{ inputs.ccud-version || steps.determine-current-version.outputs.ccud-version }}" >> "$GITHUB_OUTPUT"
+
+      - name: Log current CCUD versions
+        run: |
+          echo "dev-ccud-version = ${{ steps.determine-current-version.outputs.ccud-version }}"
+          echo "test-ccud-version = ${{ steps.determine-test-version.outputs.ccud-version }}"
+
+      - name: Set up Java
+        if: ${{ steps.determine-current-version.outputs.ccud-version == steps.determine-test-version.outputs.ccud-version }}
```
erichaagdev, line 37: A very minor thing, but you could create a boolean output in a previous step with the result of `steps.determine-current-version.outputs.ccud-version == steps.determine-test-version.outputs.ccud-version` and replace all of the subsequent occurrences.

erichaagdev (commented):
Just some questions / comments. 

I would have liked to see these tests implemented such that they could be invoked locally from Gradle, just as we did in the common injection script repository.

bigdaz:
> I would have liked to see these tests implemented such that they could be invoked locally from Gradle, just as we did in the common injection script repository.

Understood. But IIUC we didn't add test coverage directly in this repository in order to keep it "simple". So I implemented this in a workflow to minimize the impact.

.github/workflows/compatibility-check-single-version.yml
bigdaz: We don't verify the failure. We simply acknowledge that the compatibility check _might_ fail, so we don't verify success.

.github/workflows/compatibility-check-single-version.yml
bigdaz: Hmm. This looks like a typo.

bigdaz:
This workflow is useful for one-off determination of the current compatibility matrix, but perhaps is less useful on an ongoing basis.
I'll close the PR and keep my fork until we decide which direction we want to head.

PR #384, opened 2025-04-18
Add detecting and tagging VSCode builds

ribafish:
As proposed in the [discovery doc](https://docs.google.com/document/d/1bDw1n95tdx943LvXXZuAiLzfH4-o8HHuqQAdoh2oHCU/edit?tab=t.0), I've implemented supoprt for detecting and tagging builds triggered in VSCode. That includes "sync" builds, as well as regular builds, triggered via [gradle-java](https://marketplace.visualstudio.com/items?itemName=vscjava.vscode-gradle) VSCode extension.

Detection of VSCode builds is based on the `VSCODE_PID` env var, which we expect to always be set. 

I will open similar PRs for other CCUD implementations once this is approved.


src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -132,6 +134,8 @@ public void execute(BuildResultAdapter buildResult) {
                 tagIde("IntelliJ IDEA", props.get(SYSTEM_PROP_IDEA_VERSION).get());
             } else if (props.get(SYSTEM_PROP_ECLIPSE_BUILD_ID).isPresent()) {
                 tagIde("Eclipse", props.get(SYSTEM_PROP_ECLIPSE_BUILD_ID).get());
+            } else if (props.get(ENV_VAR_VSCODE_PID).isPresent()) {
+                tagIde("VSCode", "");
```
clayburn, line 138: Are we certain VSCode provides no version information that we can access?

clayburn (approved):
LGTM with one question

src/main/java/com/gradle/CustomBuildScanEnhancements.java
gabrielfeo: If we register VS Code version information, I think the ["Gradle for Java" extension](https://github.com/microsoft/vscode-gradle) version would also be relevant.

gabrielfeo, line 138: We should tag it as their [brand guidelines](https://code.visualstudio.com/brand#brand-name) ask

```suggestion
                tagIde("VS Code", "");
```

ribafish: That would all be great and all, but we don't have that info. [Here](https://ge.solutions-team.gradle.com/s/gcy6yongdfqho/console-log?task=:printout) is a scan with all the env vars, sys props, and proj props printed out.. And no version information. 

The VS Code I built this with is v 1.99.3, closest info I can find is [VSCODE_IPC_HOOK](https://ge.solutions-team.gradle.com/s/gcy6yongdfqho/console-log?task=:printout#L58), but I wouldn't base that info off a path, which can probably be changed in a future version / OS... plus, it doesn't even give us the full version. 

As for the "Gradle for Java" extension, this was built with 3.16.4, which does appear in 4 places in the above scan, but it's always in a path, which can again probably be changed.

But if you can find either the VSCode version or the ["Gradle for Java" extension](https://github.com/microsoft/vscode-gradle) version that's "stable", I'm all for it 😄 .

gabrielfeo (approved):

src/main/java/com/gradle/CustomBuildScanEnhancements.java
erichaagdev: There's also `code --list-extensions --show-versions` for extensions.

clayburn: I would also be hesitant to get Plugin versions for now. I'd like to see a better use case (beyond "more data") for how that would help a user. We would also need to consider this for other IDEs.

erichaagdev: If we don't want to call out to another process, that's fine with me (although we already do it for `git`). However, I wouldn't say `and it seems that we did check and it is not`, because it's quite easy with one command. ;)

ribafish: @erichaagdev, are you sure that `code` is always set up for cli usage?

erichaagdev: I wouldn't say "always", but I was under the impression it comes with the main app installation by default.

erichaagdev: > Visual Studio Code has a powerful command-line interface built-in that lets you control how you launch the editor. You can open files, install extensions, change the display language, and output diagnostics through command-line options (switches).

https://code.visualstudio.com/docs/configure/command-line

erichaagdev: It's confusing because it also says:

> **Things we're <ins>not</ins> ok with**
> Using the product name — Visual Studio Code or VS Code — on your website, product, service, or in your domain names.

erichaagdev: Not advocating for that though. No one is going to use `tag:Visual Studio Code`. Instinctively it will be `tag:VS Code`.

ribafish: Yes, however:

> Note: Users on macOS must first run a command (Shell Command: Install 'code' command in PATH) to add VS Code executable to the PATH environment variable. Read the [macOS setup guide](https://code.visualstudio.com/docs/setup/mac) for help.
> 
> Windows and Linux installations should add the VS Code binaries location to your system path. If this isn't the case, you can manually add the location to the Path environment variable ($PATH on Linux). For example, on Windows, the default VS Code binaries location is AppData\Local\Programs\Microsoft VS Code\bin. To review platform-specific setup instructions, see [Setup](https://code.visualstudio.com/docs/setup/setup-overview).
> 
> Insiders: If you are using the VS Code [Insiders](https://code.visualstudio.com/insiders) preview, you launch your Insiders build with `code-insiders`.

ribafish: I'm happy to add this and we can just do best effort here, if that's what we want to do. Perhaps best if we discuss it on our sync.

PR #385, opened 2025-04-18
Fixed getting git metadata when started from a different dir to the project

ribafish:
Fixed getting git data when the gradle build is started from a different directory and given a path to the repository. This can happen if the build is started with the `-p` / `--project-dir` argument, or when using the [GradleConnector](https://docs.gradle.org/current/javadoc/org/gradle/tooling/GradleConnector.html) as the main entrypoint - which is what is used by the VSCode [gradle build server](https://devblogs.microsoft.com/java/new-build-server-for-gradle/), added by [gradle-java](https://marketplace.visualstudio.com/items?itemName=vscjava.vscode-gradle) VSCode extension.

To overcome this issue, I've added the `-C <root-project-path>` argument when running git commands, which will take care of such cases.

I will investigate and open similar PRs for other CCUD implementations if we want to add this, and this is approved.

### DIFF
--- a/src/main/java/com/gradle/CommonCustomUserDataGradlePlugin.java
+++ b/src/main/java/com/gradle/CommonCustomUserDataGradlePlugin.java
@@ -69,7 +69,7 @@ public class CommonCustomUserDataGradlePlugin implements Plugin<Object> {
 
         BuildScanAdapter buildScan = develocity.getBuildScan();
         customDevelocityConfig.configureBuildScanPublishing(buildScan);
-        CustomBuildScanEnhancements buildScanEnhancements = new CustomBuildScanEnhancements(develocity, providers, settings.getGradle(), settings.getRootDir().getAbsolutePath());
+        CustomBuildScanEnhancements buildScanEnhancements = new CustomBuildScanEnhancements(develocity, providers, settings.getGradle(), settings.getRootDir());
         buildScanEnhancements.apply();
 
         BuildCacheConfiguration buildCache = settings.getBuildCache();
@@ -146,7 +146,7 @@ public class CommonCustomUserDataGradlePlugin implements Plugin<Object> {
 
         BuildScanAdapter buildScan = develocity.getBuildScan();
         customDevelocityConfig.configureBuildScanPublishing(buildScan);
-        CustomBuildScanEnhancements buildScanEnhancements = new CustomBuildScanEnhancements(develocity, providers, project.getGradle(), project.getRootDir().getAbsolutePath());
+        CustomBuildScanEnhancements buildScanEnhancements = new CustomBuildScanEnhancements(develocity, providers, project.getGradle(), project.getRootDir());
         buildScanEnhancements.apply();
 
         // Build cache configuration cannot be accessed from a project plugin
--- a/src/main/java/com/gradle/CommonCustomUserDataGradlePlugin.java
+++ b/src/main/java/com/gradle/CommonCustomUserDataGradlePlugin.java
@@ -69,7 +69,7 @@ public class CommonCustomUserDataGradlePlugin implements Plugin<Object> {
 
         BuildScanAdapter buildScan = develocity.getBuildScan();
         customDevelocityConfig.configureBuildScanPublishing(buildScan);
-        CustomBuildScanEnhancements buildScanEnhancements = new CustomBuildScanEnhancements(develocity, providers, settings.getGradle());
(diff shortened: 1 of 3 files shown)


src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -414,11 +418,11 @@ public void execute(BuildScanAdapter buildScan) {
...
-            String gitRepo = execAndGetStdOut("git", "config", "--get", "remote.origin.url");
-            String gitCommitId = execAndGetStdOut("git", "rev-parse", "--verify", "HEAD");
-            String gitCommitShortId = execAndGetStdOut("git", "rev-parse", "--short=8", "--verify", "HEAD");
-            String gitBranchName = getGitBranchName(() -> execAndGetStdOut("git", "rev-parse", "--abbrev-ref", "HEAD"));
-            String gitStatus = execAndGetStdOut("git", "status", "--porcelain");
+             String gitRepo = execAndGetStdOut("git", "-C", rootPath, "config", "--get", "remote.origin.url");
+             String gitCommitId = execAndGetStdOut("git", "-C", rootPath, "rev-parse", "--verify", "HEAD");
+             String gitCommitShortId = execAndGetStdOut("git", "-C", rootPath, "rev-parse", "--short=8", "--verify", "HEAD");
+             String gitBranchName = getGitBranchName(() -> execAndGetStdOut("git", "-C", rootPath, "rev-parse", "--abbrev-ref", "HEAD"));
+             String gitStatus = execAndGetStdOut("git", "-C", rootPath, "status", "--porcelain");
```
gabrielfeo, line 425: While it'd be a broader change, `execAndGetStdOut` could instead run these subprocesses [using the build's root dir as working directory](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/Runtime.html#exec(java.lang.String[],java.lang.String[],java.io.File)). Using `-C` is fine, I'm just pointing out an alternative.

src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -67,12 +67,14 @@ final class CustomBuildScanEnhancements {
     private final BuildScanAdapter buildScan;
     private final ProviderFactory providers;
     private final Gradle gradle;
+    private final String rootPath;
```
gabrielfeo, line 70: For consistency, could it be named `rootDir`?

gabrielfeo (commented):
Nice to learn what was causing that!

src/main/java/com/gradle/CommonCustomUserDataGradlePlugin.java
```diff
@@ -146,7 +146,7 @@ private void applyProjectPlugin(Project project, DevelocityAdapter develocity) {
 
         BuildScanAdapter buildScan = develocity.getBuildScan();
         customDevelocityConfig.configureBuildScanPublishing(buildScan);
-        CustomBuildScanEnhancements buildScanEnhancements = new CustomBuildScanEnhancements(develocity, providers, project.getGradle());
+        CustomBuildScanEnhancements buildScanEnhancements = new CustomBuildScanEnhancements(develocity, providers, project.getGradle(), project.getRootDir().getAbsolutePath());
```
clayburn, line 149: Nitpick, but stylistically, I'd lean towards passing just the rootDir here as a a file, all the way into `execAndGetStdOut` have that be the only place where we need to care about `-C`

gabrielfeo (approved):

src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -395,15 +398,17 @@ public void execute(BuildResultAdapter buildResult) {
...
     }
 
     private static final class CaptureGitMetadataAction implements Action<BuildScanAdapter> {
 
         private final ProviderFactory providers;
         private final DevelocityAdapter develocity;
+        private final File projectDir;
 
-        private CaptureGitMetadataAction(ProviderFactory providers, DevelocityAdapter develocity) {
+        private CaptureGitMetadataAction(File projectDir, ProviderFactory providers, DevelocityAdapter develocity) {
```
jprinet, line 410: I feel aligning the argument order with the constructor of `CustomBuildScanEnhancements` would help to read the code. The ordering was unaligned before your change though.

jprinet (approved):
LGTM, just some minor format related comments

jprinet (approved):
Great!

Issue #389, opened 2025-05-05
Add the job id and step information to custom values when running in Github Actions


ribafish:
Add the `job_id` and step information to custom values when running in GitHub Actions.

I believe this is valuable information for complex CI setups, with multiple jobs and multiple steps (eg, with matrices), which would make it easier to track the exact origin of a build scan in a CI pipeline. 

Checking the [default env vars](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables) for Github Actions, we can see the information available as:

- `GITHUB_ACTION`: The name of the action currently running, or the [id](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsid) of a step. For example, for an action, __repo-owner_name-of-action-repo.
- `GITHUB_JOB`: The [job_id](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_id) of the current job. For example, greeting_job.

If this is accepted, I will do the same for [CCUD Maven](https://github.com/gradle/common-custom-user-data-maven-extension) (and [CCUD sbt](https://github.com/gradle/common-custom-user-data-sbt-plugin)), where it will be especially important, since, for example, with Maven, the profiles are not tracked by default.

PR #409, opened 2025-07-08
Added setting CI Action and CI job id custom values on github action builds

ribafish:
Added setting CI Action and CI job id custom values on github action builds. [This](https://github.com/ribafish/reproducers/actions/runs/16146564695) is an example GHA builds (the `build` scan), with the [scan](https://ge.solutions-team.gradle.com/s/ykzv2537i72te) having two custom values (and search links) added: [CI job id](https://ge.solutions-team.gradle.com/s/ykzv2537i72te/custom-values#L0) and [CI step id](https://ge.solutions-team.gradle.com/s/ykzv2537i72te/custom-values#L3). This will be helpful when you've got multiple builds running consecutively and are passing in arguments that affect those builds, but are not captured in a build scan (best example are Maven profiles for CCUD Maven).

When this is accepted I will do the same for CCUD Maven (and sbt). 

Closes https://github.com/gradle/common-custom-user-data-gradle-plugin/issues/389.

### DIFF
--- a/src/main/java/com/gradle/CustomBuildScanEnhancements.java
+++ b/src/main/java/com/gradle/CustomBuildScanEnhancements.java
@@ -306,6 +306,10 @@ final class CustomBuildScanEnhancements {
                     addCustomValueAndSearchLink(develocity, "CI run", value));
                 envVariable("GITHUB_HEAD_REF", providers).filter(value -> !value.isEmpty()).ifPresent(value ->
                     buildScan.value("PR branch", value));
+                envVariable("GITHUB_ACTION", providers).ifPresent(value ->
+                        addCustomValueAndSearchLink(develocity, "CI step id", value));
+                envVariable("GITHUB_JOB", providers).ifPresent(value ->
+                        addCustomValueAndSearchLink(develocity, "CI job id", value));
             }
 
             if (isGitLab(providers)) {
--- a/src/main/java/com/gradle/CustomBuildScanEnhancements.java
+++ b/src/main/java/com/gradle/CustomBuildScanEnhancements.java
@@ -304,12 +304,12 @@ final class CustomBuildScanEnhancements {
                     addCustomValueAndSearchLink(develocity, "CI workflow", value));
                 envVariable("GITHUB_RUN_ID", providers).ifPresent(value ->
                     addCustomValueAndSearchLink(develocity, "CI run", value));
-                envVariable("GITHUB_HEAD_REF", providers).filter(value -> !value.isEmpty()).ifPresent(value ->
-                    buildScan.value("PR branch", value));
                 envVariable("GITHUB_ACTION", providers).ifPresent(value ->
-                        addCustomValueAndSearchLink(develocity, "CI step id", value));
+                        addCustomValueAndSearchLink(develocity, "CI step", value));
                 envVariable("GITHUB_JOB", providers).ifPresent(value ->
-                        addCustomValueAndSearchLink(develocity, "CI job id", value));
+                        addCustomValueAndSearchLink(develocity, "CI job", value));
+                envVariable("GITHUB_HEAD_REF", providers).filter(value -> !value.isEmpty()).ifPresent(value ->
+                        buildScan.value("PR branch", value));
             }
 
             if (isGitLab(providers)) {


src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -306,6 +306,10 @@ public void execute(BuildResultAdapter buildResult) {
                     addCustomValueAndSearchLink(develocity, "CI run", value));
                 envVariable("GITHUB_HEAD_REF", providers).filter(value -> !value.isEmpty()).ifPresent(value ->
                     buildScan.value("PR branch", value));
+                envVariable("GITHUB_ACTION", providers).ifPresent(value ->
+                        addCustomValueAndSearchLink(develocity, "CI step id", value));
+                envVariable("GITHUB_JOB", providers).ifPresent(value ->
+                        addCustomValueAndSearchLink(develocity, "CI job id", value));
```
erichaagdev, line 312: I would remove `id`.

```suggestion
                envVariable("GITHUB_ACTION", providers).ifPresent(value ->
                        addCustomValueAndSearchLink(develocity, "CI step", value));
                envVariable("GITHUB_JOB", providers).ifPresent(value ->
                        addCustomValueAndSearchLink(develocity, "CI job", value));
```

erichaagdev (changes requested):

src/main/java/com/gradle/CustomBuildScanEnhancements.java
ribafish: I'm happy to do that, however the values are only getting us `id` values, not names. If there's no id, it will show the name of whatever is being executed, if it's an external action, it'll show the name of that, for example something like `_gradle_actions_setup-gradle`, or `_run` and if there's muliple steps of the same thing, it'll show `_run_4`. I'm happy to drop the `id`, but this makes it clearer.

erichaagdev: We don't use `id` in any of the other values. Even when we capture `GITHUB_RUN_ID`, the custom value and link are named `CI run`. To me, appending `id` adds no value here and would introduce an inconsistency with all other values. 

Also keep in mind that the link gets `build scans` appended to it, so how it is now we'd have `CI job id build scans`.

src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -306,6 +306,10 @@ public void execute(BuildResultAdapter buildResult) {
                     addCustomValueAndSearchLink(develocity, "CI run", value));
                 envVariable("GITHUB_HEAD_REF", providers).filter(value -> !value.isEmpty()).ifPresent(value ->
                     buildScan.value("PR branch", value));
+                envVariable("GITHUB_ACTION", providers).ifPresent(value ->
+                        addCustomValueAndSearchLink(develocity, "CI step id", value));
```
erichaagdev, line 310: Why do we capture `GITHUB_ACTION` then call it `CI step id`? I think `CI action` would be more appropriate.

src/main/java/com/gradle/CustomBuildScanEnhancements.java
ribafish: Yes, but there it's evident it's going to be an `id`, here one could expect to get the `name` for example, which you can't.

erichaagdev: I don't necessarily disagree with you, but I still think remaining consistent and not using `id` should be the preferred.

src/main/java/com/gradle/CustomBuildScanEnhancements.java
```diff
@@ -306,6 +306,10 @@ public void execute(BuildResultAdapter buildResult) {
                     addCustomValueAndSearchLink(develocity, "CI run", value));
                 envVariable("GITHUB_HEAD_REF", providers).filter(value -> !value.isEmpty()).ifPresent(value ->
                     buildScan.value("PR branch", value));
+                envVariable("GITHUB_ACTION", providers).ifPresent(value ->
```
clayburn, line 309: Nitpick, but I'd group all of the env vars together that produce a value + link. In other words, I'd have the GITHUB_HEAD_REF logic last in this code block.

clayburn (approved):

erichaagdev (approved):

PR #410, opened 2025-07-08
Add testing latest changes using Maven Local

ribafish:
Added a Publish to Maven Local and Test with Locally Published Plugin job to Verify Build workflow. This will publish the latest changes to Maven Local, modify `settings.gradle.kts` to use the locally published plugin and run a regular build, effectively giving us a build, [built with the latest changes of CCUD Gradle plugin](https://ge.solutions-team.gradle.com/s/go5p5ozhbx3j6/build-dependencies?expandAll&focusedDependency=WzAsMSwxLFswLDEsWzFdXV0).

I implemented it by modifying the CCUD version applied to the build to be dynamic (2+). I added this to the `build-verification.yml` workflow. If you feel this makes more sense to live separately, let me know.

Once this is merged, I will add a similar thing to CCUD Maven and sbt.

### DIFF
--- a/.github/workflows/build-verification.yml
+++ b/.github/workflows/build-verification.yml
@@ -18,11 +18,18 @@ jobs:
         uses: gradle/actions/setup-gradle@v4
         with:
           develocity-access-key: ${{ secrets.DV_SOLUTIONS_ACCESS_KEY }}
-      - name: Build with Gradle
-        run: ./gradlew build -x signPluginMavenPublication -i -Porg.gradle.java.installations.auto-download=false
+      - name: Build and publish to Maven Local with Gradle
+        run: ./gradlew build publishToMavenLocal -x signPluginMavenPublication -i -Porg.gradle.java.installations.auto-download=false
+        env:
+          DISABLE_REQUIRED_SIGNING: true
+      - name: Upload published plugin
+        uses: actions/upload-artifact@v4
+        with:
+          name: common-custom-user-data-gradle-plugin
+          path: ~/.m2/repository/com/gradle
 
-  selftest:
-    name: Publish to Maven Local and Test with Locally Published Plugin
+  local-test:
+    name: Test with Locally Published Plugin
     runs-on: ubuntu-latest
     needs: verification
     steps:
@@ -37,11 +44,11 @@ jobs:
         uses: gradle/actions/setup-gradle@v4
         with:
           develocity-access-key: ${{ secrets.DV_SOLUTIONS_ACCESS_KEY }}
-      - name: Publish to maven local
-        id: publish-to-maven-local
-        run: ./gradlew clean publishToMavenLocal -Dgradle.cache.remote.push=false
-        env:
-          DISABLE_REQUIRED_SIGNING: true
+      - name: Download plugin to maven local
+        uses: actions/download-artifact@v4
+        with:
+          name: common-custom-user-data-gradle-plugin
+          path: ~/.m2/repository/com/gradle
       - name: Create a test project
         run: |
           echo """


jprinet (commented):
This certainly works 👍
I am just curious if you considered using a sample project using CCUD as an included build to avoid all the tedious publication process

.github/workflows/build-verification.yml
```diff
@@ -20,3 +20,40 @@ jobs:
...
+        uses: actions/setup-java@v4
+        with:
+          java-version: '21'
+          distribution: 'temurin'
+      - name: Set up Gradle
+        uses: gradle/actions/setup-gradle@v4
+        with:
+          develocity-access-key: ${{ secrets.DV_SOLUTIONS_ACCESS_KEY }}
+      - name: Disable signing on CI
+        run: sed -i 's/isRequired = providers.environmentVariable("CI").isPresent/isRequired = false/' build.gradle.kts
```
erichaagdev, line 41: I would avoid editing build files like this.

Instead, I recommend including a check for publishToMavenLocal as a requested task in the signing condition.

Adding a system property to disable signing if included on the command line would also be a good option.

.github/workflows/build-verification.yml
```diff
@@ -20,3 +20,40 @@ jobs:
...
+      - name: Publish to maven local
+        id: publish_to_maven_local
```
erichaagdev, line 43: I think `-` is more idiomatic for GitHub Actions.

.github/workflows/build-verification.yml
```diff
@@ -20,3 +20,40 @@ jobs:
...
+            pluginManagement {\
+            repositories {\
+            mavenLocal()\
+            gradlePluginPortal()\
+            mavenCentral()\
+          }}\
+          ' settings.gradle.kts
+          sed -i -E 's/(id\("com.gradle.common-custom-user-data-gradle-plugin"\) version )"2[0-9\.]*"/\1"2+"/' settings.gradle.kts
+
+      - name: Run a build with the locally published plugin
```
erichaagdev, line 57: Is the idea to just verify the build succeeds with CCUD applied? You don't need to run a full build to do that.

erichaagdev (changes requested):
Are you going to do the same for Maven too?

FTR, we do something similar to this for the convention samples:

https://github.com/gradle/develocity-build-config-samples/blob/main/.github/workflows/convention-develocity-gradle-plugin-verification.yml

jprinet:
> Actually, I created a [sample project](https://github.com/gradle/common-custom-user-data-gradle-plugin/compare/main...gk/sampleTest) and was ready to open a different PR...but I'm not sure I'd include a sample project, in which we define `mavenLocal` and add ccud with a dynamic version... so, while this current way is more convoluted and hacky, it's still just a test. To me, the sample project should be the way it's supposed to be used. I'm happy to add a sample project, but I still think it shouldn't have `mavenLocal` and dynamic versions in it.

That's why I would recommend building the sample project with CCUD plugin as an included build, you would get it applied without the need to publish to Maven local, nor specifying the version

ribafish:
> I don't think we should confuse sample projects and projects used for testing. I also don't think we should use included builds as it's not how the plugin is consumed by users.
> 
> What about writing out a minimal build file to a temporary directory and running the test there? Something like this is all you need:
> 
> Then run `gradle help`.

Good idea, probably the least convoluted and clearest among the discussed. I've implemented it, [this is the run with it](https://github.com/gradle/common-custom-user-data-gradle-plugin/actions/runs/16194157806).

erichaagdev (approved):
Approved, but left a few suggestions.

PR #419, opened 2025-08-12
Gradle 9.0 upgrade and fixes

ribafish:
Builds on/replaces https://github.com/gradle/common-custom-user-data-gradle-plugin/pull/413, with fixes necessary because `forUseAtConfigurationTime` was removed in Gradle 9.0 - access to it is now moved to reflection.

That is still necessary when the CCUD plugin is applied to projects with Gradle versions between 6.0 (inclusive) and 7.4 (exclusive), as starting with version 7.4 Gradle will implicitly treat an external value used at configuration time as a configuration cache input. More info [here](https://docs.gradle.org/9.0.0/userguide/upgrading_version_7.html#for_use_at_configuration_time_deprecation).

### DIFF
--- a/src/main/java/com/gradle/Utils.java
+++ b/src/main/java/com/gradle/Utils.java
@@ -50,7 +50,7 @@ public final class Utils {
 
     static Optional<String> envVariable(String name, ProviderFactory providers) {
         if (isGradle65OrNewer() && !isGradle74OrNewer()) {
-            Provider<String> variable = reflectionForUseAtConfigurationTime(providers.environmentVariable(name));
+            Provider<String> variable = forUseAtConfigurationTime(providers.environmentVariable(name));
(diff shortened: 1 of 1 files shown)


src/main/java/com/gradle/Utils.java
```diff
@@ -266,6 +267,20 @@ private static Optional<URI> toUri(String scheme, String host, String path) {
         }
     }
 
+    private static <T> Provider<T> reflectionForUseAtConfigurationTime(Provider<T> provider) {
```
erichaagdev, line 270: minor: I would just call it `forUseAtConfigurationTime`. The fact that it uses reflection is an irrelevant implementation detail.

erichaagdev, line 270: Can it just return `Provider<String>`?

erichaagdev (approved):

PR #432, opened 2025-09-15
Remove references to enterprise-samples.gradle.com

guylabs:
We want to decommission enterprise-samples.gradle.com and therefore we remove all references to it.


jprinet (approved):
LGTM

PR #513, opened 2026-06-23
Introduce MAP_AGENT_FINGERPRINTS_ENV to support more AI Agents

ganadist:
Support more agents
  * Antigravity
  * Github Copilot
  * Cline
  * Cursor 
  
Adjusted to detect Gemini in Android Studio


erichaagdev:
@ganadist – thanks for the contribution, and for adding the original agent detection that this builds on.

As a team, we've decided to be more selective about which AI tools we detect going forward, so we won't be taking all of these. We'd also rather keep the structure of the existing file than move to the map approach, and maintain backward compatibility with the `android.studio.agent` property, so I found it easier to recreate the pieces we wanted directly.

I've given you kudos in the new PR.

PR #537, opened 2026-08-05
Migrate CI to the Develocity Artifact Cache GitHub Action

jdpgrailsdev:
## What

Migrates this repo's CI (`build-verification.yml` and `build-verification-nightly.yml`) from the hand-rolled Develocity Artifact Cache CLI setup — a custom "setup + restore" composite action plus a separate "store" action, driven by `DV_SERVER` / `ARTIFACT_CACHE_*` environment variables — to the first-party **[Develocity Artifact Cache GitHub Action](https://github.com/gradle/develocity-artifact-cache-github-action)**.

Each cache integration collapses to a single step:

```yaml
- uses: gradle/develocity-artifact-cache-github-action@v0
  with:
    develocity-url: https://ge.solutions-team.gradle.com
    develocity-access-key: ${{ secrets.DV_SOLUTIONS_ACCESS_KEY }}
    ac-cli-version: 1.5.0
    additional-cli-args: --gradle-home=/home/runner/.gradle
```

The action restores in its main step and stores automatically in a post step, so the separate store step and the custom composite actions are removed.

## Notes

- **Targets `@v0`** — the current published major of the action. It is pre-GA; the tag will become `@v1` at general availability, at which point this reference should be bumped.
- **`ac-cli-version: 1.5.0` is pinned** deliberately, matching the CLI version this CI used before, so the migration is apples-to-apples rather than also changing the CLI.
- This is **dogfooding ahead of the action's GA**.

## Draft: CI is expected to fail until the action is public

The action's distribution repository (`gradle/develocity-artifact-cache-github-action`) is currently **internal**, so this **public** repo's workflows cannot resolve `uses: …@v0` yet — the action step will fail until that repo is made public (or org access is granted). Holding this as a **draft** until then; please do not merge before the action is generally available.

### DIFF
--- a/.github/workflows/build-verification-nightly.yml
+++ b/.github/workflows/build-verification-nightly.yml
@@ -10,17 +10,24 @@ on:
       - ".github/workflows/build-verification-nightly.yml"
   workflow_dispatch:
 env:
+  # --- MIGRATION: superseded by the Develocity Artifact Cache GitHub Action.
+  #     These vars only supported the custom CLI actions; values reused in the
+  #     action's `with:` block on the migrated steps below.
   # Artifact Cache CLI Configuration
-  ARTIFACT_CACHE_CLI_VERSION: 1.2.0
+  # ARTIFACT_CACHE_CLI_VERSION: 1.2.0    -> action input: ac-cli-version
+  #   (bumped to 1.5.0, the latest AC CLI, as part of this migration)
+  #
+  ## Develocity
+  # DV_SERVER: https://ge.solutions-team.gradle.com   -> action input: develocity-url
+  #
+  ## Misc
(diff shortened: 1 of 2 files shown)


erichaagdev:
@jdpgrailsdev – I see and understand the draft disclaimer, but is the PR ready to be reviewed aside from that?

jdpgrailsdev:
> @jdpgrailsdev – I see and understand the draft disclaimer, but is the PR ready to be reviewed aside from that?

@erichaagdev It is, but ideally, I would be granted push access to the actual repo instead of having to use a fork so that we can see the action actually work/run on as part of the PR checks.

jdpgrailsdev:
Superseded by #538, which moves the branch onto the repo itself so its CI receives the Develocity access-key secret and actually exercises the cache (a fork PR does not get secrets). Closing this one.

PR #538, opened 2026-08-05
Migrate CI to the Develocity Artifact Cache GitHub Action

jdpgrailsdev:
## What

Migrates this repo's CI (`build-verification.yml` and `build-verification-nightly.yml`) from the hand-rolled Develocity Artifact Cache CLI setup — a custom "setup + restore" composite action plus a separate "store" action, driven by `DV_SERVER` / `ARTIFACT_CACHE_*` environment variables — to the first-party **[Develocity Artifact Cache GitHub Action](https://github.com/gradle/develocity-artifact-cache-github-action)**.

Each cache integration collapses to a single step:

```yaml
- uses: gradle/develocity-artifact-cache-github-action@v0
  with:
    develocity-url: https://ge.solutions-team.gradle.com
    develocity-access-key: ${{ secrets.DV_SOLUTIONS_ACCESS_KEY }}
    cli-version: 1.5.0
    additional-cli-args: --gradle-home=/home/runner/.gradle
```

The action restores in its main step and stores automatically in a post step, so the separate store step and the custom composite actions are removed.

## Notes

- **Targets `@v0`** — the current published major of the action. It is pre-GA; bump to `@v1` at general availability.
- **`cli-version: 1.5.0` is pinned** deliberately, matching the CLI version this CI used before, so the migration is apples-to-apples rather than also changing the CLI.
- This is **dogfooding ahead of the action's GA**. Unlike the earlier fork-based attempt, this PR runs on a branch in the repo itself, so its CI receives `DV_SOLUTIONS_ACCESS_KEY` and exercises real cache restore/store.

Supersedes the fork-based draft #537.


jdpgrailsdev:
Put back to draft status.  I am going to continue to use this PR to test pre-releases of the action.  Once we have a stable pre-release/GA release, we can flip this back to ready for review.

jdpgrailsdev:
PR has been updated to use the latest RC of the GitHub Action.  I separate PR will be raised once the GA is cut, which will largely be an identical bundle to the RC.

.github/workflows/build-verification-nightly.yml
```diff
@@ -10,18 +10,9 @@ on:
       - ".github/workflows/build-verification-nightly.yml"
   workflow_dispatch:
 env:
-  # Artifact Cache CLI Configuration
-  ARTIFACT_CACHE_CLI_VERSION: 1.2.0
-
-  ## Develocity
-  DV_SERVER: https://ge.solutions-team.gradle.com
+  # Also used for Build Scan publishing, not only the Artifact Cache action.
```
clayburn: I don't think this comment adds much. I'd remove it.

.github/workflows/build-verification-nightly.yml
```diff
@@ -40,17 +31,17 @@ jobs:
...
           develocity-access-key: ${{ secrets.DV_SOLUTIONS_ACCESS_KEY }}
           gradle-version: 'release-candidate'
-      - name: Setup and restore Universal Cache
-        uses: ./.github/actions/setup-and-restore-artifact-cache
+      - name: Develocity Artifact Cache
+        uses: gradle/develocity-artifact-cache-github-action@v0.0.5-rc-3
+        with:
+          develocity-url: https://ge.solutions-team.gradle.com
+          develocity-access-key: ${{ secrets.DV_SOLUTIONS_ACCESS_KEY }}
+          cli-version: 1.5.0
```
clayburn, line 38: @jdpgrailsdev - this is needed for now, yes? We can't automatically get the latest version compatible with our Develocity instance?

.github/workflows/build-verification-nightly.yml
```diff
@@ -40,17 +31,17 @@ jobs:
...
+          additional-cli-args: --gradle-home=/home/runner/.gradle
```
clayburn: I assume this is the GitHub Actions default Gradle home. If so, is it possible for this to be the default such that we don't have to specify it? Maybe there is a good reason for requiring it.

.github/workflows/build-verification-nightly.yml
```diff
@@ -80,8 +71,17 @@ jobs:
           cache-disabled: true
           develocity-access-key: ${{ secrets.DV_SOLUTIONS_ACCESS_KEY }}
           gradle-version: '${{ matrix.gradle-version }}'
-      - name: Setup and restore Universal Cache
-        uses: ./.github/actions/setup-and-restore-artifact-cache
+      # `matrix` is passed so the action warns that these gradle-version legs
+      # share one auto-generated cache. See build-verification.yml for the
+      # image-names pattern that isolates each gradle-version.
+      - name: Develocity Artifact Cache
```
clayburn: Preference, but I'd prefer the name remain an imperative clause, such as `Setup Develocity Artifact Cache`, to match the other step names

.github/workflows/build-verification-nightly.yml
```diff
@@ -80,8 +71,17 @@ jobs:
...
+        uses: gradle/develocity-artifact-cache-github-action@v0.0.5-rc-3
+        with:
+          develocity-url: https://ge.solutions-team.gradle.com
+          develocity-access-key: ${{ secrets.DV_SOLUTIONS_ACCESS_KEY }}
+          cli-version: 1.5.0
+          additional-cli-args: --gradle-home=/home/runner/.gradle
+          matrix: ${{ toJSON(matrix) }}
```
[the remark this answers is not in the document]
clayburn: @jdpgrailsdev - I understand how the matrix is handled in the build-verification.yml, but I'm confused by its handling here and why it is different here, than there (especially in light of this Copilot comment)

.github/workflows/build-verification-nightly.yml
jdpgrailsdev: @clayburn Not needed unless you want to control when you bump to the latest.  If you don't care about picking up the latest, which could be incompatible with the Edge, I can remove this.

.github/workflows/build-verification-nightly.yml
jdpgrailsdev: @clayburn This is a AC CLI thing, not the action.  I would have to check what the AC CLI defaults to for the gradle home.  If it matches this, we can avoid this.  However, I believe your custom actions were explicitly passing this to the CLI, which implies that it is or was not the default before.

.github/workflows/build-verification-nightly.yml
jdpgrailsdev: Done

.github/workflows/build-verification-nightly.yml
jdpgrailsdev: Done

.github/workflows/build-verification-nightly.yml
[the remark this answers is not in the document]
jdpgrailsdev: This is a left over miss from when we removed the matrix parameter from the action, but still need to use the version to ensure names don't collide.  I have remove the erroneous matrix parameter and added the suggestion from copilot, which is correct.

.github/workflows/build-verification-nightly.yml
jdpgrailsdev: @clayburn Sorry...I misread your question.  The Edge is about to start serving the CLI.  However, the action doesn't integrate with that yet.  For now, if you omit the version, you get the latest version that the action knows about.  Once we integrate with the Edge serving, you would then get the latest one that edge/DV supports.

.github/workflows/build-verification-nightly.yml
jdpgrailsdev: @clayburn On a second pass, this has been removed.  I forgot that we set the new action to honor Gradle user home env var exported by setup-gradle, so setting it is redundant.

.github/workflows/build-verification-nightly.yml
clayburn: Got it, let's leave this in for now then.

clayburn (approved):

.github/workflows/build-verification-nightly.yml
```diff
@@ -80,8 +69,16 @@ jobs:
           cache-disabled: true
           develocity-access-key: ${{ secrets.DV_SOLUTIONS_ACCESS_KEY }}
           gradle-version: '${{ matrix.gradle-version }}'
-      - name: Setup and restore Universal Cache
-        uses: ./.github/actions/setup-and-restore-artifact-cache
+      # These legs differ only by gradle-version, which the auto-generated cache
+      # name does not isolate, so a shared cache image would thrash on save
+      # (last writer wins). image-names gives each gradle-version its own image.
```
erichaagdev, line 74: I recommend dropping this comment. What it refers to is already quite self-explanatory and a common pattern in matrix workflows.

There's a similar comment in the `build-verification.yml` file.

jdpgrailsdev: @erichaagdev I'll raise a PR for this.

PR #545, opened 2026-08-21
Remove the image-names cache-isolation rationale comment

jdpgrailsdev:
## Summary

In response to https://github.com/gradle/common-custom-user-data-gradle-plugin/pull/538#discussion_r3825093178

Removes the multi-line comment above the **Setup Develocity Artifact Cache** step in both `build-verification.yml` and `build-verification-nightly.yml` that explained why each Gradle-version matrix leg needs its own `image-names` (to avoid a shared cache image thrashing on save). The `image-names: ccud-gradle[-nightly]-${{ matrix.gradle-version }}` value is self-explanatory, so the rationale block is dropped.

No behavior change — comment-only removal; the `image-names` inputs and every other step are untouched.

## Files
- `.github/workflows/build-verification.yml` — drop 6-line comment
- `.github/workflows/build-verification-nightly.yml` — drop 3-line comment

### DIFF
--- a/.github/workflows/build-verification-nightly.yml
+++ b/.github/workflows/build-verification-nightly.yml
@@ -69,9 +69,6 @@ jobs:
           cache-disabled: true
           develocity-access-key: ${{ secrets.DV_SOLUTIONS_ACCESS_KEY }}
           gradle-version: '${{ matrix.gradle-version }}'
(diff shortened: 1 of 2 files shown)


gabrielfeo (approved):

erichaagdev:
Thanks!

PR #555, opened 2026-09-14
Capture all build scan metadata at execution time

### LINKED ISSUE #552: Develocity and this plugin should avoid reading env variables during configuration time
Develocity and this plugin should avoid reading env variables during configuration time. As far as I can tell these are only needed for reporting and not actual configuration. As is, any change to these env variables will invalidate configuration cache. Ideally you read these env variables later or via providers API.

Image

erichaagdev:
Every environment variable and system property the plugin captures as build scan data is now read at execution time, so none of them become configuration cache inputs. The only reads left at configuration time are the `Overrides` settings that configure Develocity and the build cache, which have to invalidate the cache when they change.

CI detection was the main cause. It reads up to 14 environment variables and runs during configuration when the plugin is applied, registering them all as inputs, so changing any of them invalidates the cache even though none of them affect the configuration.

The Git metadata background action was a second, narrower cause on Gradle 6.5 through 7.3. `Utils` served both the configuration-time reads in `Overrides` and the execution-time reads everywhere else with a single implementation. On those versions, the only way to get a tracked read is `forUseAtConfigurationTime`, which registers the value whenever it is queried rather than only during configuration, so the background action's reads were tracked despite never running during configuration. Reads are now split by phase, and `Overrides` keeps the configuration inputs it actually needs.

Resolves #552


etiennestuder:
@erichaagdev When a build fails at configuration time, will all custom tags/values/links previously captured at config-time no longer be captured?

